PR #30, opened 2024-02-22
First pass at the SMDX 2.0 transform function

RedWalters:
### What

Hardcoded elements replaced with arguments (at this point they're input path, output path). Alongside a function to generate the supporting metadata with similar changes (arguments being path to transformed csv, path to supporting structure xml, output path, path to metadata template, and optional argument for possible config file).

To note this code is kind of a mess, and currently no one who might have an inclination whether the output of the transformation fits the requirements has looked at it so there's a good chance a bunch of changes will need to be made.

### How to review

Attempt to run the code locally only passing the function the arguments required and see if it breaks. Assuming I didn't miss anything it shouldn't complain. Also to note, don't try and use a full SMDX file while testing this, 'cause the full files take like 5 minutes to run. I need to investigate whether this is due to my really dodgy code or whether it's just 'cause the files are rather large. My assumption is that's it's kinda both. 

Any notes on improving the code itself would also be welcome, though not a requirement.

### Who can review

Only I worked on this so anyone should be able to review this.


mikeAdamss (commented):
added a couple of minor comments. not gonna review transform or approve pr, there's not me folks who should know how this works 😄

pipelines/pipeline/functions/transform/SDMX_2_0.py
```diff
@@ -205,7 +205,7 @@ def generate_editions_metadata(transformedCSV, structureXML, outputPath, metadat
 
     # dump out metadata file
 
-    with open(outputPath + pathify(dataset_title) + "_" + str(datetime.now().strftime("%Y-%m")) + "-metadata.json", "w") as outfile:
+    with open(outputPath + pathify_func(dataset_title) + "_" + str(datetime.now().strftime("%Y-%m")) + "-metadata.json", "w") as outfile:
         json.dump(editions_template, outfile, ensure_ascii=False, indent=4)
```
Shannon95, line 211: All looks fine with changes made to SDMX_2_0.py as seems to be change of naming conventions and brining in the pathify_function. Only aspect that might need addressed is comment mike has made above.

Shannon95 (commented):
Changes made look fine to me. Although I believe there are a few previous comments made by Mike that need addressing but from chatting to J he will "_be implementing the feedback and requested changes into a new PR later_"

PR #32, opened 2024-02-23
Pipeline Notification

bailed22:


mikeAdamss (commented):
Re dependencies (slack client), the _intended_ behaviour is:

- we `pipenv install git+https://github.com/ONSdigital/dp-python-tools.git` to this codebase so that slack client is available same as any other 3rd party python package you might choose to use.
- then you'll be able to just import it like it was pandas or any other tool your code chooses to use. 
- .....but.... we're not quite there yet, which I suspect made this a tad more challenging that it might otherwise have been 😄 

Now, Sarah is having a look at sorting that out, so:

- tidy up all the other bits on this PR, do that assertions bit I asked for etc
- put a pin in it and do something else if the slack client isn't migrated by the time you do that.
- once the slack client is migrated - install dpytools and make use of it then this one will be ready for PR again.

one side point - its worth getting in the habit of double checking what files and changes you've actually put into PR on github before posting it in slack, I'm pretty sure there's some files in here you didn't mean to add 😄

mikeAdamss (commented):
Looking good 👍 , couple of minor comments and we need to test the assertions are being raised where the env vars have not been specified.

put your test in `./tests/pipelines/pipeline/shared/test_notify.py`

this is an example test for it, you'll need one per env var:

```python
import pytest

def test_notify_se_webook_raises_for_missing_env_var():
    """
    Test that when we import notify, an assertion error
    is correctly raised if we're missing the env var
    specifying the software engineering slack webhook
    """

    # use monkeypatch to set all the env vars other than
    # the one you're testing for the absense of
    mp = pytest.MonkeyPatch()
    mp.setenv("PS_SLACK_WEBHOOK", "not-used")
    mp.setenv("DE_SLACK_WEBHOOK", "not-used")
    mp.setenv("DS_SLACK_WEBHOOK", "not-used")

    with pytest.raises(AssertionError) as e:
        from pipelines.pipeline.shared import notify

    assert "Unable to find required environment variable SE_SLACK_WEBHOOK" in str(e)
```

hope that makes sense.

mikeAdamss (changes requested):
trivial change needed (doc string in tests) then happy to approve 😄

mikeAdamss (approved):
LGTM

Issue #13, opened 2024-02-06
validate csv


mikeAdamss:
## What is this

We need a function that can validate the contents of a csv file. This should use the accompanying metdata file to do this (the metadata file should have relevant details - such as what columnw should be present).

## What to do

**NOTE** - metadata structure is still in flux at time of writing, if we don't have a solid metadata structure at time of writing just populate the `_` methods. You'll need to use some common sense here, do as much as is feasible with the information we have at the time.

I'm imagining something like:

```python

# Intended to be called from function at bottom
def _correct_columns_exist(csv, columns: List[str]):
    """
    given a list of columns, make sure those columns are in the csv
    """
    ...

# Intended to be called from function at bottom
def _column_has_type(csv, column: str, type: Any):
    """
    given a csv a column name and a type, make sure that all
    values in that column can be case to that type (int, float, str etc)
    """
    ...

# Intended to be called from function at bottom
def _column_has_no_blanks(csv, column: str):
    """
    given a csv a column name conform that there are
    no blank entries in the column.
    """
    ...

# Main validation function, intended for being called from the pipeline
def validate_csv(csv_path: Path, metadata_path: Path):
    """
    Validate a csv
    """
    # Assert the csv_path points to a csv
    # Qssert the metadata_path points to a json file
    # Make sure both can be read in and throw useful errors if
    # they cannot.

    # the methods above (and other to follow) will need to follow
    # once we have more information.
    ...
```

Which would raise an Exception in the event of any issue.

Note - I'm not a DE. If someone in this space can see a cleaner way of doing this then that's cool, but remember it needs to be modular. Lots of little functions are very easy to unit test, the more monolithic a thing the harder that gets.

If you do use pandas please load only what you need (so a single column if we're just checking a columns content, just the headers if checking the name etc). Be wary of dumping everyting into a data frame, "everything" could be a very large file indeed. If we can partially insulate ourselves against that by only reading in what we need then that'd be ideal.  

## Acceptance Criteria

- functions written to the extent we currently can
- unit tests for first three functions

mikeAdamss:
done

PR #52, opened 2024-03-07
validate csv and test

JimBryant7:
### What

Added validate_csv.py given from ticket #13 which is a number of functions that will validate the outputted csv against a few factors - checking data against expected data given from metadata & making sure files can be read in and contain no empty values etc
There was no location given to store this file so have made an educated guess on where it should sit (one level up from where the transform will be) happy to be corrected on this.
Also contains unit tests for the functions as well as 2 basic (very trimmed down version) fixtures - a csv file and a metadata file, these are based on the current output that I have and may need updating in the future

### How to review

Mainly a sanity check, also check that the unit tests pass

### Who can review

Anyone


mikeAdamss (commented):
So you've done exactly what I asked but in hindsight I think I got it wrong (sorry 😄 ).

The _idea_ (mitigating against large datasets) was ok, but my per-column approach is off.

What I _think_ (double check my thinking) is that we should not validate _per column_ but instead validate a _slice of rows_, i.e

- work out the number of rows in the csv (use a csv reader or something tht doesnt require loading the whole thing)
- read X (lets say 2000) rows
- validate them
- validate the next block of rows

That way each function just needs to be a thing that operates against a dataframe which I think will be easier for people to udnerstand (also, my original thought would just break if someone gives us a massive column, SAPE timeseries etc)

This way for small datasets (most) it's just validating a dataframe, the complexity only comes in when we get larger that typical datasets (which is kinda what we want).

We do also have the start of the new metadata structure kicking around somewhere - if you could make start on the above I'll see if I can get you a copy of that for you to work to.

mikeAdamss:
So there are several outstanding questions around how we handle dimensions that need to be bottomed out before we can get into any validation that references the metadata structure.

Given that, can you:

- delete the overarching `validate_csv` function (for now)
- create a `dataframing_slicing` funtion of some kind - just a function that _could_ be used by the above function when we implement it.
- build out the other functions as stand alones that take a dataframe
- tests for everything

basically, dont get into the metadata structure at all for this, we'll pick that up later - make this into create the "functions that validate dataframes" toolbox and we'll kick putting it all together to a different task.

pipelines/pipeline/shared/transforms/validate_csv.py
```diff
@@ -0,0 +1,95 @@
+# this is a "functions that validate dataframes" toolbox
+# will be pieced together in another ticket (delete this text when it gets picked up)
+# more functions can be added as needed
+
+
+from typing import List
+from pathlib import Path
+import pandas as pd
+
+def _read_in_csv_check(csv_path):
```
mikeAdamss: can you type hint this please, `csv_path: Path`

pipelines/pipeline/shared/transforms/validate_csv.py
```diff
@@ -0,0 +1,95 @@
...
+    """
+    given a csv will read in first 5 lines as a check
+    """
+    assert csv_path.name.endswith('.csv'), "Invalid csv_path"
+    
+    try:
+        pd.read_csv(csv_path, nrows=5)
+    except:
```
mikeAdamss: if you do this as

```python
except Exception as err:
     raise Exception(f"Failed to read in cs. - {csv_path}") from err
```

then your new error gets added at the top of the stack trace but the orignal error is in there as well (the "this error was cause by this other error..." thing python does).

pipelines/pipeline/shared/transforms/validate_csv.py
```diff
@@ -0,0 +1,95 @@
...
+        raise Exception(f"Failed to read in csv - {csv_path}")
+        
+        
+def _correct_columns_exist(csv_path, columns: List[str]):
```
mikeAdamss: type hint, `csv_path: Path`

pipelines/pipeline/shared/transforms/validate_csv.py
```diff
@@ -0,0 +1,95 @@
...
+    assert len(df.columns) == len(columns), f"Number of columns in csv file ({len(df.columns)}) does not match len of expected columns ({len(columns)})"
+    
+    for col in columns:
+        assert col in df.columns, f"Expect column {col} not found in csv file"
+        
+    # number of columns are the same and all expected columns are in df so no
+    # need to check the other way around
+
+
+def _dataframe_has_no_blanks(df):
```
mikeAdamss: can you add a keyword argument of say `check_specific_columns: Optional[List] = None)`

that way, if its None we check all columns, but retain the freedom to specify. I'm thinking of things of things like attributes/observation level metadata. I don't know if we'll have them, but need to account for a scenario where it might be _valid_ for certain columns to have blanks.

pipelines/pipeline/shared/transforms/validate_csv.py
```diff
@@ -0,0 +1,95 @@
...
+    """
+    given a df and a column name confirm that there are
+    no blank entries in the column.
```
mikeAdamss: "in the specified columns, or all columns where "check_specific_columns" is None.

pipelines/pipeline/shared/transforms/validate_csv.py
```diff
@@ -0,0 +1,95 @@
...
+    """
+    for col in df.columns:
+        # will treat '' & pd.isnull as blanks                 
+        for value in df[col].unique():
+            if type(value) == bool:
+                # bool can only be True of False so not empty
+                continue
+            elif type(value) == str:
+                if not value.strip(): # catches whitespace strings
+                    raise ValueError(f"{col} has blank entries - '{value}' found")
```
mikeAdamss: value would be "" if I'm following. Would probably be a confusing error. Just giving the column name will do

pipelines/pipeline/shared/transforms/validate_csv.py
```diff
@@ -0,0 +1,95 @@
...
+    """
+    assert len(df) == len(df.drop_duplicates()), "Found duplicate rows in the dataframe, failed validation"
+    
+
+def dataframing_slicing(csv_path) -> List[pd.core.frame.DataFrame]:
+    """
+    given a csv path read in the dataframe as slices and return these slices as a list
+    """
+    # currently a generic slice size, theres probably a most efficient size
+    chunk_size = 5000
```
mikeAdamss: can we pass it in as a keyword argument with this as the default as well please, might be handy to allow people to change the default if they need to.

pipelines/pipeline/shared/transforms/validate_csv.py
```diff
@@ -0,0 +1,95 @@
...
+                    
+
+def _dataframe_has_no_duplicates(df):
+    """
+    given a df confirm that there are no duplicate values
+    """
+    assert len(df) == len(df.drop_duplicates()), "Found duplicate rows in the dataframe, failed validation"
+    
+
+def dataframing_slicing(csv_path) -> List[pd.core.frame.DataFrame]:
```
mikeAdamss: so this one gets really interesting 😈 

we want to avoid putting the whole thing in memory at once if we can, so the logic needs to be something like:

- rename to `generated_dataframe_slices` or similar.
- count the rows without using a dataframe
- work out something like a dict of rows tracking the start and end of each slice that we want, so its its a 10020 rows csv, we'd want `{0:, 5000, 5000: 10000, 10000: 10021}`
- iterate that dict and for each one yields a dataframe consisting of _only_ that slice of rows (there should be some way of doing that via pd.read_csv). You'll need to make sure that each one has the correct header row too.

what you;re aiming for is to be able to do something like:

```python
for df in generated_dataframe_slices(csv_path):
      # run checks against this dataframe that is a subset of rows
      # this subset is the only thing pandas has in memory
```

which _looks_ simple but the magic trick here is that because it a generator (yield) at any given time _only_ the one subset of rows is in memory, so it'll work against absolutely **massive** datasets without hitting the pandas resource limits.

_note - cant remember if you can iterate a generator in quite that fashion, you. might need to finagle it, but the ideas sound._

pipelines/pipeline/shared/transforms/validate_csv.py
mikeAdamss: type hint that df is pd.DataFrame as well please.

mikeAdamss (commented):
added some comments, give me a shout if you need some clarified.

tests/pipelines/pipeline/shared/transforms/test_validate_csv.py
```diff
@@ -0,0 +1,106 @@
...
+        _dataframe_has_no_duplicates(df)
+        
+    assert "Found duplicate rows in the dataframe, failed validation" in str(err.value)
+
+def test_generated_dataframe_slices():
+    fixture_file = Path(fixtures_files_dir / 'test_validate_csv_data.csv')
+
+    try:
+        for df in generated_dataframe_slices(fixture_file, chunk_size=2):
+            assert type(df) == pd.DataFrame, "generated_dataframe_slices not returning a pd.DataFrame"
```
mikeAdamss, line 110: you should really test that its returning a dataset with a number of rows equal to what you've asked for as well

mikeAdamss (commented):
one minor thing then happy to approve

mikeAdamss (commented):
looks good but need a little more test coverage for the genertator.

like start with a csv of 13 rows and read in 5 rows at a time, confirm you get the 5,,5 then 3 df's and that each has the correct number of rows and the expected headers or something of that ilk. this sort of thing can be fiddly, need to strongly enforce we're not clipping off a row anywhere

mikeAdamss (approved):
LGTM

PR #54, opened 2024-03-07
Minor tidy up

mikeAdamss:
### What

Correcting some minor things I've missed at pr.

- makefile to use new package
- ran `make fmt`
- ran `make lint` and removed an unused import i found
- pin the python version to some variation of 3.9 (for compatibility with aws glue)

### How to review

Sanity check, confirm you can `make` things.

### Who can review

Anyone.


pyproject.toml
```diff
@@ -7,7 +7,7 @@ readme = "README.md"
 include = ["dpypelines/schemas/**/*.json"]
 
 [tool.poetry.dependencies]
-python = "3.9"
+python = ">=3.9.0,<3.10.0"
```
Moasib-Arif, line 10: I noticed that in order to run 'Make test' or any other command, I had to delete the existing 'poetry.lock' file because I encountered this error: 'pyproject.toml changed significantly since poetry.lock was last generated. Run poetry lock [--no-update] to fix the lock file.' 

The previous 'poetry.lock' file was configured for Python 3.9, while the updates included changes to the Python version specified in the 'pyproject.toml' file to '3.9.0,<3.10.0'.

mikeAdamss: good spot 😄 , have regenerated the lock file

Moasib-Arif (approved):
LGTM

PR #55, opened 2024-03-08
initial function to take an s3 object name

mikeAdamss:
### What

Update start function to take an s3 object name as the positional argument.

This is needed to hook this codebase into the infrastructure (which will call this function with said object name).

### How to review

Sanity check

### Who can review

Anyone


Moasib-Arif (approved):
LGTM

PR #56, opened 2024-03-11
Create messages 2

NickPapONS:
### What

Adds 3 new functions for error messages and unit tests for each one checking the returned message contents.

### How to review

Check that the error messages are sensible and contain reasonable details on the errors. Check that the tests are making the right assertions/checks.

Please let me know what your thoughts are on my use of local_path in the message text. I could not think of a way to use the input store otherwise to add detail to the message, but local_path depends on `LocalDirectoryStore` which is an implementation of `BaseWritableDirectoryStore`. If I shouldn't use that here I am happy to change it.

### Who can review

Anyone.

### DIFF
--- a/pipelines/pipeline/shared/message.py
+++ b/pipelines/pipeline/shared/message.py
@@ -1,0 +1,154 @@
+
+"""
+Functions to create nicely formatted and informative messages.
+"""
+from os import pipe
+from pathlib import Path
+import json
+from typing import Dict
+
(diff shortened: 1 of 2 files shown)


SarahJohnsonONS (approved):
LGTM - messages look sensible, tests all passing. Re `local_path`, you could use the `get_current_source_pathlike` abstract method, which would mean it wouldn't matter if it wasn't called `local_path` in a different type of store. It looks like you might need to do a Poetry update so that the data-pipelines version of dpytools has that method implemented.

PR #58, opened 2024-03-12
S3 tar recieved

nimshi89:
### What

added a check for the decompressing the s3_tar file
added a check for a config file existing
added a check to confirm if the config file is valid

Can't run the pipeline yet 

### How to review

supply the tar file in the bucket  by start("tarfile_path")
then when the pipleline will be working that will run as well

### Who can review

anyone


dpypelines/s3_tar_received.py
```diff
@@ -1,22 +1,53 @@
...
     """
+    try:
+        # IN ALL CASES, the FOLLOWING STEPS WILL BE IN TRY CATCH
+        # BLOCKS AND WILL NOTIFY RELEVANT PARTIES IN EVENT OF FAILURE
 
-    # IN ALL CASES, the FOLLOWING STEPS WILL BE IN TRY CATCH
-    # BLOCKS AND WILL NOTIFY RELEVANT PARTIES IN EVENT OF FAILURE
+        # decompress tar file to workspace
+        files_path = Path("./")
+        decompress_s3_tar(s3_object_name, files_path)
```
mikeAdamss: you don't need a path object and I would just use a simple and obviously named directory (the code will creata it), so 

```python
decompress_s3_tar(s3_object_name, "input")
```

will do

dpypelines/s3_tar_received.py
```diff
@@ -1,22 +1,53 @@
...
+        localStore = LocalDirectoryStore(files_path)
+    except Exception as error:
+        notify.data_engineering("ERROR MESSAGE")
+        raise error
 
-    # decompress tar file to workspace
+        # confirm we have a config
+    try:
+        if not localStore.has_lone_file_matching("config.json"):
```
mikeAdamss: assume its called `pipeline-config.json`, that's what we were going for.

dpypelines/s3_tar_received.py
```diff
@@ -1,22 +1,53 @@
...
+        notify.data_engineering("ERROR MESSAGE")
+        raise error
 
-    # confirm we have a config
+        # if we dont have a config - make one? (tbd)
 
-    # if we dont have a config - make one? (tbd)
+        # use config schema to confirm that config is valid
+    try:
+        config_dict = localStore.get_lone_matching_json_as_dict("path_to_config_file")
```
mikeAdamss: `pipeline-config.json` not `path_to_confif_file`

dpypelines/s3_tar_received.py
```diff
@@ -1,22 +1,53 @@
...
+    try:
+        # IN ALL CASES, the FOLLOWING STEPS WILL BE IN TRY CATCH
+        # BLOCKS AND WILL NOTIFY RELEVANT PARTIES IN EVENT OF FAILURE
 
-    # IN ALL CASES, the FOLLOWING STEPS WILL BE IN TRY CATCH
-    # BLOCKS AND WILL NOTIFY RELEVANT PARTIES IN EVENT OF FAILURE
+        # decompress tar file to workspace
+        files_path = Path("./")
+        decompress_s3_tar(s3_object_name, files_path)
+        localStore = LocalDirectoryStore(files_path)
```
mikeAdamss: as above, just use `LocalDirectoryStore("input")`

mikeAdamss (commented):
Minor comments aside your approach is right 😄 

one thing, don't put bits of executable logic in a single try catch (you've got a try around both decompress and making local store). Put them in separate try catch blocks please - we're aiming fora. big ol verbose page with every bit of logic wrapped in a try catch with a notification.

by the same logic line 42, 43 and 44 should all be in separate catches.

mikeAdamss:
Had to do something very similar to check the IAC was hooking up correctly, have included here -  just follow this sort of pattern and continue with the extra bits that are on this task.

```python
from dpytools.s3.basic import decompress_s3_tar
import os

from dpytools.stores.directory.local import LocalDirectoryStore

from dpypelines.pipeline.shared import notification as notify

def start(s3_object_name: str):
    """
    Handles the required behaviour when recieving a tar file indicated
    by an s3 object name.

    Example s3_object_name: my-bucket/my-data.tar
    """

    # Decompress the tar to ./inputs
    try:
        decompress_s3_tar(s3_object_name, "inputs")
    except Exception as err:
        notify.data_engineering("MESSAGE")
        raise err

    # Identify the child directory we've just decompressed to within /inputs
    try:
        directories = [d for d in os.listdir("inputs") if os.path.isdir(os.path.join("inputs", d))]
        assert len(directories) == 1, (
            "Aborting, input directory has more than one directory within it"
        )
        decompressed_directory = directories[0]
    except Exception as err:
        notify.data_engineering("MESSAGE")
        raise err

    # Create a local directory store using our new files
    try:
        store = LocalDirectoryStore(f"inputs/{decompressed_directory}")
    except Exception as err:
        notify.data_engineering("MESSAGE")
        raise err
```

dpypelines/s3_tar_received.py
```diff
@@ -1,22 +1,103 @@
...
-    # IN ALL CASES, the FOLLOWING STEPS WILL BE IN TRY CATCH
-    # BLOCKS AND WILL NOTIFY RELEVANT PARTIES IN EVENT OF FAILURE
+    # Decompress the tar file to the workspace
+    try:
+        decompress_s3_tar(s3_object_name, "input")
+    except Exception as err:
+        notify.data_engineering(message.unexpected_error(f"Failed to decompress tar file {s3_object_name}", err))
+        raise err
+    
+    # Identify the child directory that was just decompressed within /inputs
```
mikeAdamss: so this - it turned out I was needing to do this because the tar file was created in an odd manner, so a folder was compressed to create it rather than just compressing some files into an archive at the root level.

so - we're just going to insist that there's no sub folder. The input files can then be decompressed directly into "./inputs".

so this can go, as can the reference to `{decompressed_directory}`.

it worked, but I don't want to get into handling sub folders, we'll just require that people provide things in the way we want (i.e not in a sub folder).

dpypelines/s3_tar_received.py
```diff
@@ -1,22 +1,103 @@
...
+        notify.data_engineering(message.unexpected_error(f"Failed to create local directory store at inputs/{decompressed_directory}", err))
+        raise err
+
+    # Verify if a pipeline configuration file ('pipeline-config.json') exists in the local directory store
+    try:
+        # Check for the existence of a configuration file
+        if not localStore.has_lone_file_matching("pipeline-config.json"):
+            msg = message.expected_local_file_missing("Missing pipeline-config.json", "pipeline-config.json", "start")
+            notify.data_engineering(msg)
+            # Create a default configuration file if one does not exist
```
mikeAdamss: dont do this, I know there was a tbd (to be decided) but its still not decided. Just assume there is one and raise if there's not.

if we tackle this, it'll be as a separate piece later.

for now, just raise an exception here.

dpypelines/s3_tar_received.py
```diff
@@ -1,22 +1,103 @@
...
+            default_config = {
+                "pipeline": "default_pipeline",
+                "parameters": {}
+            }
+            with open("pipeline-config.json", "w") as f:
+                json.dump(default_config, f)
+        else:
```
mikeAdamss: you can get rid of the else and just unindent 63 once we're raising above

dpypelines/s3_tar_received.py
```diff
@@ -1,22 +1,103 @@
...
+    try:
+        path_to_schema = get_config_schema_path(config_dict)
+    except Exception as err:
+        notify.data_engineering(message.cant_find_schema(config_dict, err))
+        raise err
 
-    # confirm we have a config
+    # Validate the configuration against the retrieved schema
+    try:    
+        validation.validate_json_schema(schema_path=path_to_schema, data_dict=config_dict)
```
mikeAdamss: have a look at the signature for `validate_json_schema` you should be able to pass a message, i.e "validating pipeline-config.json" and `indent=2`. That will make any raised exceptions a lot easier on the eye.

dpypelines/s3_tar_received.py
```diff
@@ -1,22 +1,103 @@
...
+    except Exception as err:
+        notify.data_engineering(message.invalid_config(config_dict, err))
+        raise err
 
-    # if we dont have a config - make one? (tbd)
+    # Loads the pipeline specified in the configuration file and executes it with the provided parameters
```
mikeAdamss: clever but don't worry about this just yet please (there's tasks to come on this). For now just explicitly call: https://github.com/ONSdigital/dp-data-pipelines/blob/0682f4e8ca72fc5309bbd032c56a85acceeedb92/dpypelines/pipeline/dataset_ingress_v1.py#L9 with the path to the directory (there's a method on the store to get it)

mikeAdamss (commented):
Well done. A few comments but there all simlifications 😄

dpypelines/s3_tar_received.py
```diff
@@ -1,22 +1,75 @@
...
+        localStore = LocalDirectoryStore(f"inputs")
+    except Exception as err:
+        notify.data_engineering(message.unexpected_error(f"Failed to create local directory store at inputs", err))
+        raise err
 
-    # decompress tar file to workspace
+    # Check for the existence of a configuration file
+    try:
+        if not localStore.has_lone_file_matching("pipeline-config.json"):
+            msg = message.expected_local_file_missing("Missing pipeline-config.json", "pipeline-config.json", "start")
```
mikeAdamss: can you just use `message.unexpected_error()`, this message doesn't quite match up as I initially thought it would. We'll have a task to look at this coming up.

dpypelines/s3_tar_received.py
```diff
@@ -1,22 +1,75 @@
...
+    except Exception as err:
+        notify.data_engineering(message.unexpected_error("Error while checking for pipeline-config.json", err))
+        raise err
 
-    # confirm we have a config
+    # Load the configuration file and validate it against a schema
+    try:
+        config_dict = localStore.get_lone_matching_json_as_dict("pipeline-config.json")
+    except Exception as err:
+        notify.data_engineering(message.pipeline_input_exception({"pipeline-config.json": "expected"}, localStore, err))
```
mikeAdamss: just use `message.unexpected_error()` for now please. I need to do some cleaning up on messages.

dpypelines/s3_tar_received.py
```diff
@@ -1,22 +1,75 @@
...
+        raise err
 
-    # use config schema to confirm that config is valid
+    # if we dont have a config - make one? (tbd)
 
-    # use the config info to select and run the appropriate pipeline,
-    # there should literally by a field in the pipeline-config telling
-    # us which pipeline from ./pipeline/* to call.
+    # Run the dataset_ingress_v1 pipeline
+    try:
```
mikeAdamss: its counter intuitive but don't bother with the catch on this last one 😄 

everything in `data_ingress_v1()` has its own catch block with notification, having a second more generic notification will just make it harder to grep whats going on.

mikeAdamss (commented):
last couple of minor points

dpypelines/s3_tar_received.py
```diff
@@ -1,22 +1,67 @@
...
+    try:    
+        validation.validate_json_schema(schema_path=path_to_schema, data_dict=config_dict, error_msg="Validating pipeline-config.json", indent=2)
+    except Exception as err:
+        notify.data_engineering(message.invalid_config(config_dict, err))
+        raise err
 
-    # use the config info to select and run the appropriate pipeline,
-    # there should literally by a field in the pipeline-config telling
-    # us which pipeline from ./pipeline/* to call.
+    # if we dont have a config - make one? (tbd)
```
mikeAdamss: getting wires crossed, can you just call `https://github.com/ONSdigital/dp-data-pipelines/blob/0682f4e8ca72fc5309bbd032c56a85acceeedb92/dpypelines/pipeline/dataset_ingress_v1.py#L9` here to finish this script please.

to explain, the logic for this initial script is _intended_ (long term vision) to be:

- 1.) check there's a config
- 2.) use the config to call the _appropriate_ secondary script

its to account for the fact that (over time) a tar file and a submitted dataset wont necessarily always be the same thing.

you've done 1 and we're _not_ gonna do anything clever with 2 just yet (as we don't need to, there's only one option for the next step to call) so just explicitly call it for now (without a try catch please)

mikeAdamss (commented):
one last line and we're done

dpypelines/s3_tar_received.py
```diff
@@ -1,22 +1,70 @@
...
+        validation.validate_json_schema(schema_path=path_to_schema, data_dict=config_dict, error_msg="Validating pipeline-config.json", indent=2)
+    except Exception as err:
+        notify.data_engineering(message.invalid_config(config_dict, err))
+        raise err
 
-    # use the config info to select and run the appropriate pipeline,
-    # there should literally by a field in the pipeline-config telling
-    # us which pipeline from ./pipeline/* to call.
+    # Get the path to the directory
+    files_dir = localStore.local_path
```
mikeAdamss: can you use `get_current_source_pathlike()` please, that returns a string path to a directory (which is what we need) and will work for all implementations of directory store. local_path is a `Path` object.

i wouldn't block a pr for this next one on its own, but just noticed it, its `local_store` or `store` not `localStore` throughout please. snake case not camel case.

mikeAdamss (commented):
one comment

mikeAdamss (approved):
lgtm 😄

PR #60, opened 2024-03-12
#quick fixing file structure

NickPapONS:
Some files were accidentally left over in a directory we don't want to exist anymore. Now moving them to preferred locations and cleaning up unnecessary folders.


mikeAdamss (approved):
LGTM 👍

PR #67, opened 2024-03-19
lock down regex, add more representative transform namespacing

mikeAdamss:
### What

Another consolidation pr to tidy things us.

- Added a "succeeded this far" type message to confirm the pipeline logic is running successfully to the implemented stage.
- Tightened up the regex (will explain below)
- fixed a message generator that was missing an argument
- added a suitable namespace to identify the transform after a chat with JJ.

RE regex - the `r` just means "raw string" so allows you to not care about escaping special characters (like '.'). `^` means start of string, `$` means end of string. Without this it can annoyingly match unexpected system files generated during the decompression (i.e it was double matching and finding a file called `_pipeline-config.json` which we don't care about).

### How to review

Set the 4 env vars from here: https://github.com/ONSdigital/dp-data-pipelines/blob/sandbox/dpypelines/pipeline/shared/notification.py to the webhook (see slack channel or ask in slack for a webhook). literally `export DE_SLACK_WEBHOOK=<the webhook url` etc etc in your terminal.

run this in a script in the root of the repo:

```python
from dpypelines import s3_tar_received

s3_tar_received.start("dp-bleed-ingest-submission-bucket/valid.tar")
```

you'll probably need to be signed into aws via SSO to do this. If it works you should get an "it worled" type message via a notification to slack (if it doesn't you should get error messages via slack too)

### Who can review

Anyone with aws SSO.

### DIFF
--- a/dpypelines/s3_tar_received.py
+++ b/dpypelines/s3_tar_received.py
@@ -43,9 +43,10 @@ def start(s3_object_name: str):
 
     # Check for the existence of a configuration file
     try:
-        if not local_store.has_lone_file_matching("pipeline-config.json"):
+        if not local_store.has_lone_file_matching(r"^pipeline-config.json$"):
+            err = FileNotFoundError(f"Cannot fine pipeline-config.json, from {local_store.get_file_names()}")
             notify.data_engineering(
-                message.unexpected_error("Missing pipeline-config.json")
+                message.unexpected_error("Cannot fine pipeline config", err)
             )
     except Exception as err:
         notify.data_engineering(
@@ -57,7 +58,7 @@ def start(s3_object_name: str):
 
     # Load the configuration file and validate it against a schema
     try:
-        config_dict = local_store.get_lone_matching_json_as_dict("pipeline-config.json")
+        config_dict = local_store.get_lone_matching_json_as_dict(r"^pipeline-config.json$")
     except Exception as err:
         notify.data_engineering(
             message.unexpected_error("Error while getting pipeline-config.json", err)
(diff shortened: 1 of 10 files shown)


dpypelines/s3_tar_received.py
```diff
@@ -43,9 +43,10 @@ def start(s3_object_name: str):
 
     # Check for the existence of a configuration file
     try:
-        if not local_store.has_lone_file_matching("pipeline-config.json"):
+        if not local_store.has_lone_file_matching(r"^pipeline-config.json$"):
+            err = FileNotFoundError(f"Cannot fine pipeline-config.json, from {local_store.get_file_names()}")
             notify.data_engineering(
-                message.unexpected_error("Missing pipeline-config.json")
+                message.unexpected_error("Cannot fine pipeline config", err)
```
Moasib-Arif: minor typo "find"

dpypelines/s3_tar_received.py
```diff
@@ -43,9 +43,10 @@ def start(s3_object_name: str):
 
     # Check for the existence of a configuration file
     try:
-        if not local_store.has_lone_file_matching("pipeline-config.json"):
+        if not local_store.has_lone_file_matching(r"^pipeline-config.json$"):
+            err = FileNotFoundError(f"Cannot fine pipeline-config.json, from {local_store.get_file_names()}")
```
Moasib-Arif: minor typo "find"

Moasib-Arif (commented):
Everything else looks good to me 👍

Moasib-Arif (approved):
LGTM

PR #68, opened 2024-03-19
63_decorator

nimshi89:
### What

added a utility file with a decorator that can be added to the messages.

### How to review

check the decorator and the test to see if it ads the correct message.

### Who can review

anyone

### DIFF
--- a/dpypelines/pipeline/shared/utility.py
+++ b/dpypelines/pipeline/shared/utility.py
@@ -5,7 +5,7 @@ def enrich_online(message_creating_function):
         msg: str = message_creating_function(*args)
         
         #Get the enviormentl variable
-        enrich_message = os.environ.get('ENRICH_OUTGOING_MESSAGES')
+        enrich_message = os.environ.get('ENRICH_OUTGOING_MESSAGES', None)
         #If there is a value stored add (enrich) message otherwise return original message
         if enrich_message is not None:
             return msg + enrich_message + "\n notification-source: ENRICH_OUTGOING_MESSAGES"
--- a/dpypelines/pipeline/shared/utility.py
+++ b/dpypelines/pipeline/shared/utility.py
@@ -8,7 +8,7 @@ def enrich_online(message_creating_function):
         enrich_message = os.environ.get('ENRICH_OUTGOING_MESSAGES', None)
         #If there is a value stored add (enrich) message otherwise return original message
         if enrich_message is not None:
-            return msg + enrich_message + "\n notification-source: ENRICH_OUTGOING_MESSAGES"
+            return msg + enrich_message
         else:
              return msg
         
--- a/dpypelines/pipeline/shared/utility.py
+++ b/dpypelines/pipeline/shared/utility.py
@@ -1,8 +1,8 @@
 import os
 
 def enrich_online(message_creating_function):
-    def wrapper(**kwargs):
-        msg: str = message_creating_function(**kwargs)
+    def wrapper(*args, **kwargs):
+        msg: str = message_creating_function(*args, **kwargs)
         
         #Get the enviormentl variable
         enrich_message = os.environ.get('ENRICH_OUTGOING_MESSAGES', None)
--- a/dpypelines/pipeline/shared/utility.py
+++ b/dpypelines/pipeline/shared/utility.py
@@ -1,14 +1,14 @@
 import os
 
 def enrich_online(message_creating_function):
-    def wrapper(*args):
-        msg: str = message_creating_function(*args)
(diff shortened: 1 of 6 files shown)


dpypelines/pipeline/shared/utility.py
```diff
@@ -0,0 +1,15 @@
...
+
+def enrich_online(message_creating_function):
+    def wrapper(*args):
+        msg: str = message_creating_function(*args)
+        
+        #Get the enviormentl variable
+        enrich_message = os.environ.get('ENRICH_MESSAGE')
+        #If there is a value stored add (enrich) message otherwise return original message
+        if enrich_message is not None:
+            return msg + " " + enrich_message + "\n notification-source"
```
SarahJohnsonONS: I don't think it needs to be this complicated. If you just return `msg + enrich_message`, and then pass `\n notification-source` as the env var in the test, the original message is returned with the env var text included.

tests/pipelines/pipeline/shared/test_utility.py
```diff
@@ -0,0 +1,38 @@
+import pytest
+import os
+
+from dpypelines.pipeline.shared.utility import enrich_online
+from dpypelines.pipeline.shared.message import unexpected_error
+
+
+@enrich_online
+def very_much_expected_error(msg: str, error: Exception) -> str:
```
SarahJohnsonONS: Assuming this is here just for PR purposes - once the decorator is ready, remove and decorate the functions in `dpypelines.pipeline.shared.message`?

nimshi89: Yes I only put this there for the PR so it can be tested.

dpypelines/pipeline/shared/utility.py
nimshi89: Screenshot 2024-03-20 at 09 47 42
 did you mean  it like this?

nimshi89: Screenshot 2024-03-20 at 09 48 27
Did you mean is like this? Sorry just the second was confusing for me :)

tests/pipelines/pipeline/shared/test_utility.py
```diff
@@ -0,0 +1,38 @@
...
+
+        Error type: {error_type}
+        Error: {error}
+    """
+
+    return message
+
+
+def test_enrich_message():
+    os.environ["ENRICH_MESSAGE"] = "Please work"
```
mikeAdamss: so this will set an environment variable on the users machine until they next shut it down, needs to just be for the length of this test.

see here: https://github.com/ONSdigital/dp-python-tools/blob/b256a30a4c8d4acf4584b336f22b642c8aa933a4/tests/config/test_config.py#L16 for an example of how to set it for the length of the test only.

mikeAdamss: it doesn't need to be an exception, fundementally your decorator wraps something that returns a sting and _sometimes_ (if an env var is set) adds a bit to the end, thats all,, so we only need something very lightweight here.

```python

def test_enrich_decorator():
    """
    Test we can control the behaviour of the enrich online decorator via an env var.
    """

    @enrich_online
    def _foo_funct() -> str:
          return "foo"
          
     # no env var set
     assert _foo_func() == "foo"
     
     # TODO - set env var with monkeypatch
     
     # env var is set
     assert _foo_func() == "foo <whatever should be added"
```

dpypelines/pipeline/shared/utility.py
```diff
@@ -0,0 +1,15 @@
+import os
+
+def enrich_online(message_creating_function):
+    def wrapper(*args):
```
mikeAdamss: on this and the following line, can you pass through **kwargs as well please, of we're not already its a matter of time before we have a message function that uses a keyword or two.

dpypelines/pipeline/shared/utility.py
```diff
@@ -0,0 +1,15 @@
...
+        msg: str = message_creating_function(*args)
+        
+        #Get the enviormentl variable
+        enrich_message = os.environ.get('ENRICH_OUTGOING_MESSAGES', None)
+        #If there is a value stored add (enrich) message otherwise return original message
+        if enrich_message is not None:
+            return msg + enrich_message
```
mikeAdamss: can you add a line break between the original and added message please. It'll be a url so fresh line will probably be more readible.

mikeAdamss (commented):
minot comment on testing, as Sarah said don't forget to decorate all the messages as part of this.

Moasib-Arif (approved):
LGTM 👍

PR #69, opened 2024-03-19
JJ/DIS-566

RedWalters:
### What

~~Includes most of the changes requested issue-15 related to the main sdmx transform. But the~~ main thing to be reviewed here is the metadata generation function

### How to review

There's gonna be 2 steps to this one, the first is gonna be a review of the code itself; its mostly just pulling in and manipulating a dictionary with as much information as possible but there may be better ways to do what needs to be done. The other is a review of the actual output itself; I'm not 100% on the structure/contents so both of these things will need reviewing and possibly correcting/informing.

### Who can review

For the first part anyone really can review this, though people with experience messing with json files would be best. 
the latter will need someone familiar with the application profile so they can have some back and forth on the structure/contents of the output metadata

### DIFF
--- a/dpypelines/pipeline/functions/utils.py
+++ b/dpypelines/pipeline/functions/utils.py
@@ -1,0 +1,19 @@
+import string
+from unidecode import unidecode
+import re
+
+def pathify(label):
+    """
+      Convert a label into something that can be used in a URI path segment.
+    """
+    return re.sub(r'-$', '',
+                  re.sub(r'-+', '-',
+                         re.sub(r'[^\w/]', '-', unidecode(label).lower())))
+
+def set_key(dictionary, key, value):
+     if key not in dictionary:
+         dictionary[key] = value
+     elif type(dictionary[key]) == list:
+         dictionary[key].append(value)
+     else:
+         dictionary[key] = [dictionary[key], value]
--- a/dpypelines/pipeline/shared/transforms/sdmx/SDMX_2_0.py
+++ b/dpypelines/pipeline/shared/transforms/sdmx/SDMX_2_0.py
@@ -1,0 +1,211 @@
+import json
+import xmltodict
+import pandas as pd
+from datetime import datetime, date
+from dpypelines.pipeline.functions.utils import pathify
+
+def set_key(dictionary, key, value):
+     if key not in dictionary:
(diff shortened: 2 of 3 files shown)


mikeAdamss (commented):
so to link this into the pipeline it'll need a function with a signature of:

```python
def jjs_code(path_to_input_1: Path) -> Tuple[Path, Path]:
      # all of the stuff happens here
      
      return path_to_csv_file, path_to_metadata_json_file
```

There's a simple hard coded one here that Jim did a while ago to help is hook things together on the SE side: https://github.com/ONSdigital/dp-data-pipelines/blob/3c3653ae96794616fae5686c5ab74bb3ca1517c9/dpypelines/pipeline/shared/transforms/sdmx/v1.py#L40

Not gonna get involved in transform code, but we need that pattern (or to wrap you code un a function with that pattern) to make it all work together.

RedWalters:
> so to link this into the pipeline it'll need a function with a signature of:
> 
> ```python
> def jjs_code(path_to_input_1: Path) -> Tuple[Path, Path]:
>       # all of the stuff happens here
>       
>       return path_to_csv_file, path_to_metadata_json_file
> ```
> 
> There's a simple hard coded one here that Jim did a while ago to help is hook things together on the SE side:
> 
> https://github.com/ONSdigital/dp-data-pipelines/blob/3c3653ae96794616fae5686c5ab74bb3ca1517c9/dpypelines/pipeline/shared/transforms/sdmx/v1.py#L40
> 
> Not gonna get involved in transform code, but we need that pattern (or to wrap you code un a function with that pattern) to make it all work together.

So my plan for this was to have the functions for the transforms kept in a separate file (so this one for instance) and the have the pipeline reference these function from elsewhere which has both the correct signatures, switches for various smdx versioning and validation and what-have-you without having all the code in one place. 

So we'll have a file that looks very similar to [this](https://github.com/ONSdigital/dp-data-pipelines/blob/sandbox/dpypelines/pipeline/shared/transforms/sdmx/v1.py) with all the lengthy transform code held somewhere else that can be edited/added to as needed without risking messing with the pipeline stuff itself.

That was my idea anyway

RedWalters:
This issue is gonna be covered by issue-87

PR #71, opened 2024-03-20
Acceptance tests data fixtures

NickPapONS:
### What
Adds some functionality to create basic data fixtures (for now containing random dummy data) that can be extracted from an archive to be used for acceptance tests. The data is taken from a zipfile if it doesn't exist, then the relevant files requested by the test's context are placed in a temporary directory which is then used for the remainder of the test.

### How to review
Check the before_all functionality that ensures the directory with the fixture data inside exists and see if it makes sense.
Check data.py and see if the code structure makes sense and fulfills its purpose. There's a lot of iterating/matching things in a dictionary going on which can get a bit complicated. Let me know if you feel it needs some edge case coverage/error catching somewhere.

### Who can review

Anyone


features/fixtures/data/pipeline_config_basic_valid.json
```diff
@@ -0,0 +1,18 @@
+{
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    "$id": "https://raw.githubusercontent.com/ONSdigital/dp-data-pipelines/sandbox/schemas/dataset-ingress/config/v1.json",
+    "required_files": [
+        {
+            "matches": "*.sdmx",
```
mikeAdamss: your problem is my bad regex guess. Just match these to for example `^data.xml$` (I've seen example where the DEs are using xml over sdmx, but check with the data engineers). Same for supplementary distributions. 

`*.` is invalid, wrong way around.

features/steps/data.py
```diff
@@ -0,0 +1,55 @@
...
+
+                    save_path = temporary_dir / files_to_retrieve[file]
+                    with open(save_path, "w") as f:
+                        f.write(file_data)
+
+        context.temporary_dir = temporary_dir
+
+    # If anything goes wrong, make sure the temporary directory gets removed.
+    except Exception:
+        shutil.rmtree(temporary_dir)
```
mikeAdamss: good that we're removing but we can handle this in envionrment.py.

- in `before_sceanario` set `content.temporary_directory` to None.
- add an `after_scenario` check and if its not None, delete it.

features/fixtures/data/data_sdmx_valid_1.sdmx
mikeAdamss, line 1: should this just be in the zip file? we probably want to add `/fixtures/data` to gitignore to avoid these getting scooped up into git.

features/fixtures/data/pipeline_config_basic_valid.json
mikeAdamss, line 1: should this just be in the zip file? we probably want to add `/fixtures/data` to gitignore to avoid these getting scooped up into git.

mikeAdamss (commented):
added a few comments but this is pretty slick thus far 😄

features/fixtures/data/data_sdmx_valid_1.sdmx
SarahJohnsonONS: Have added to gitignore, but also set it to delete the `features/fixtures/data` path in `environment.after_all()`

features/fixtures/data/pipeline_config_basic_valid.json
SarahJohnsonONS: Have updated the json fixture to use `^data.sdmx$` for both `required_files` and `supplementary_distributions` (although there aren't any supplementary distributions in the unzipped folder, so I've set the `count` as 0).

`dataset_ingress_v1` now gets past the regex checking stage (but fails on a subsequent `try... except` block).

mikeAdamss:
pushed some changes as this one was a little snaggled. Two main issues going on.

1.) We were catching errors a little early, it was obfiiscating some pathing issues.

2.) We updates the pipeline for a new config shape a few pr's ago, needed to update the fixture config to reflect that, for context a valid config looks like this now:


```json
{
    "$schema": "http://json-schema.org/draft-04/schema#",
    "$id": "https://raw.githubusercontent.com/ONSdigital/dp-data-pipelines/sandbox/schemas/dataset-ingress/config/v1.json",
    "required_files": [
        {
            "matches": "^data.sdmx$",
            "count": "1"
        }
    ],
    "supplementary_distributions": [
        {
            "matches": "^data.sdmx$",
            "count": "1"
        }
    ],
    "priority": "1",
    "pipeline": "dataset_ingress_v1",
    "options": {
        "transform_identifier": "sdmx.compact.v2.0.prototype.1"
    }
}
``` 

I also added a "no expections" and "an exception" step, mainly to help check the above was working, plus a few minor code changes.

Moasib-Arif (approved):
LGTM 👍

PR #81, opened 2024-03-26
Quick Assertion error fix

Moasib-Arif:
### What

The error message in the assertion was not matching the actual error message when the FileNotFoundError is raised and the test was failing

### How to review

Run all the tests (Make Test)

### Who can review

Anyone

### DIFF
--- a/dpypelines/pipeline/shared/schemas.py
+++ b/dpypelines/pipeline/shared/schemas.py
@@ -2,7 +2,6 @@ import json
 import os
 from pathlib import Path
 
-
 def get_config_schema_path(config: dict) -> Path:
     """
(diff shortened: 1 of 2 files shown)


SarahJohnsonONS (commented):
This was failing because `schema_base_path = Path("schemas/dataset-ingress/config")` was missing part of the path - if you change it to `schema_base_path = dpypelines_absolute_path / "schemas/dataset-ingress/config"` it raises the FileNotFoundError with the correct message details. Also need to change `local_schema_path = (dpypelines_absolute_path / schema_base_path / config_schema_version)` to `local_schema_path = schema_base_path / config_schema_version`

SarahJohnsonONS (approved):
LGTM

PR #82, opened 2024-03-26
dont reference variable after failed assignment

mikeAdamss:
### What

correcting an oversight, just a simple logical error:

- in the event we could not create the variable "transform_identifier"
- we raised an error which **included** that variable we don't have

### How to review

Sanity check

### Who can review

Anyone


Moasib-Arif (approved):
LGTM 👍

Issue #49, opened 2024-03-06
toggle off notifications for local development


mikeAdamss:
## What is this

As we develop this package we'll want to check that the notifications are firing off as expected - but we probably don't want devlopment work spamming slack notifications everywhere, it'll get a bit overwhelming.

- We (probably - investigating other solutions is 100% fine) need to add an environment variable to toggle notifications off.

## What to do

And an env var that will turn notifications off, `DISABLE_NOTIFICATIONS=true` or similar (anything sensible will do). 

Currently if they're not toggled off we need to set 3-4 env vars, the logic needs to be either or.

- scenario 1 - notifications are disabled. `DISABLE_NOTIFICATIONS=true` therefore the env vars for all the notification webhooks do not need to be set.
- scenario 2 - notifications are **not** disabled. All the required env vars for notifications (the webhook urls) need to be set.

**Don't** worry about sending the message elsewhere or capturing it and putting it to local. Bonus points if you can do it but it's not 100% necessary, the fact the code wanted to send a notification will be obvious to a command line user once we've added logging.

Be careful with pythond treatment of env vars, "true" in an env var is just a string and the truty value of a non blank string is True, so a **string** of "false" has a boolean value of true. Google python env vars and str_to_bool (or similar, it's a known and solved problem anyway, you just need to account for it).

## Acceptance Criteria

- [ ] Notifications can be toggled off.

mikeAdamss:
done

PR #94, opened 2024-04-04
Disable notifications

mikeAdamss:
### What

Allow a developer to disable notifications to slack while developing by using the `DISABLE_NOTIFICATIONS` env var. This task: https://github.com/ONSdigital/dp-data-pipelines/issues/49

I've changed the nomanclature to "Notifier" over "Messenger" as its just pass/fail notifications now.

I've created a BasePipelineNotifier base class with two implementations:

- PipelineNotifier - the normal behaviour
- NopNotifier (NOP - non operational). Which has empty methods so will satisfy the code while not actually sending any notifications.

There;'s a function which decides which one you've instantiated based on the `DISABLE_NOTIFICATIONS` env var.

Ran `fmt` and `lint` to tidy up the codebase so there's quite a lot of noise here. To stop this from happening all the time I've added a github action to run everything on push to branches (for now, we should really be using concourse).

### How to review

Try it out. Sanity check.

### Who can review

Anyone.


.github/workflows/code_quality.yaml
```diff
@@ -0,0 +1,30 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.9'
+
+      - name: Install pipenv
```
SarahJohnsonONS: `name` should be `Install poetry`?

SarahJohnsonONS (commented):
LGTM - just one comment

.github/workflows/code_quality.yaml
mikeAdamss: good spot, me and my copy paste

SarahJohnsonONS (approved):
Approved

Issue #84, opened 2024-03-27
initial pipeline transform acceptance test


mikeAdamss:
**this is a knowldge transfer task and should be done by both a DE and an SE.**

## What is this

We have written some accptance tests setup that allows you to specify a feature that starts with some amount of files in a directory,

see [this](https://github.com/ONSdigital/dp-data-pipelines/blob/50846d7d1b8c14a4cced4349144d4b83066bdd02/features/data_ingress_v1.feature#L9).

We need to create some acceptance tests for the inputs we're reciveing that expand on this to check that the data is being correctly generated. 

## What to do

Now we have a transform  We need to update that example to something like this

(note - im saying cpih as I dont know the datasets, just pick one)

```yaml
  Scenario: CPIH - Pipeline runs without errors
    Given a temporary source directory of files
        | file                   |  fixture                           |
        | pipeline-config.json   |  pipeline_config_basic_valid.json  |
        | data.xml               |  data_sdmx_valid_1.sdmx            |
    And v1_data_ingress starts using the temporary source directory
    Then the pipeline should generate no errors
    And a read the csv output "data.csv"
    And the csv output should have "100" rows
    And the csv output has the columns
        | Column 1 | Column 2 | Column 3 |
    And I read the metadata output "/outputs/metadata.json"
    And the metadata should match "cpih-metadata-correct.json"
```

Note 1 - start with running the existing holding features and go from there.

Note 2 - the files inside the "fixture" column in the above are pulled from `/features/fixtures/data-fixtures.zip`. To add more choice for fixtures, unzip it -> add your file(s) - > rezip it. this is to avoid adding massive amoutns of files to the repo, by the same token try and use quite small examples if you can.

Note 3 - Currently, calling the "v1_data_ingress starts using the temporary source directory" step causes some files to be output (`data.xml`, `data.csv` and `metadata.json`). These should be output to a temporary directory, which should be deleted once the acceptance tests have finished running.

## Acceptance Criteria

- [ ] A running initial acceptance test

PR #95, opened 2024-04-08
#84 initial acceptance test

NickPapONS:
### What

This implements more of the acceptance test steps based on the groundwork done in the acceptance tests fixture ticket: https://github.com/ONSdigital/dp-data-pipelines/pull/71

It adds steps that test the outputs generated by v1_data_ingress, mainly the csv and metadata.json output files.

Another significant change in this is the new use of a temporary directory in the behaviour tests. This is created at the start of a scenario, and is used to accommodate the outputting of files during tests, to allow them to be placed in a separate directory which can be safely deleted after each scenario is done. (The output files are often placed in the "current directory", which previously was simply the root directory and so it would be difficult to delete the files there.)

### How to review

Try the behaviour tests and see if they pass in your environment. Look through the steps and see if we missed something or if something could be done better.

### Who can review

Anyone other than Shannon and Nick (both worked on this)


features/environment.py
```diff
@@ -85,13 +96,18 @@ def after_scenario(context, scenario):
     if context.temporary_directory is not None:
         shutil.rmtree(context.temporary_directory)
 
+    # Change out of temporary directory and delete its contents from finished scenario
+    os.chdir("..")
```
mikeAdamss: be careful with relative path changes, you can probably get this via `os.chdir(context.text._test_dir.parent)` or similar.

features/environment.py
```diff
@@ -59,6 +63,13 @@ def before_scenario(context, scenario):
     """
     Test setup that runs at the start of each individual scenario.
     """
+
+    # Change directory to a temporary directory to accomodate any test files being dumped to the "current directory"
+    # This allows us to safely delete them later.
+    context.temp_test_dir = Path("temporary_test_output/")
```
mikeAdamss: careful with relative paths, I'd be tempted. to write this so its not dependent on your current working directory. i.e `Path(Path(__file__).parent.parent / "temporary_output_directory")`

mikeAdamss: that'll make your temp dir always appear in the root of the repo as `temporary_output_directory` regardless of where you run the tests from.

features/steps/data.py
```diff
@@ -46,6 +48,48 @@ def step_impl(context):
         raise context.exception
 
 
+@then("I read the csv output '{csv_output}'")
+def step_impl(context, csv_output):
+    try:
+        context.csv_output = pd.read_csv(csv_output)
+
+    except Exception as exc:
```
mikeAdamss: you. don't need to catch this, you only ever do that if you need a point if intervention to do something before the error is raised which we're not doing here.

just remove.the try catch and the error will just raise of its own accord.

features/steps/data.py
```diff
@@ -46,6 +48,48 @@ def step_impl(context):
...
+
+        context.json_output = json.load(metadata_file)
+
+    except Exception as exc:
+        raise exc
+
+
+@then("the metadata should match 'cpih-metadata-correct.json'")
+def step_impl(context):
+    expected_metadata = {"to": "do"}
```
mikeAdamss: use a path to a fixture please, just so someone can change the input later

features/steps/data.py
```diff
@@ -46,6 +48,48 @@ def step_impl(context):
...
+
+    except Exception as exc:
+        raise exc
+
+
+@then("the csv output should have '{number}' rows")
+def step_impl(context, number):
+    num_rows = len(context.csv_output.index)
+
+    assert num_rows == int(number)
```
mikeAdamss: can you add a message to be shown if the assertion fails, i.e

```python
assert num_rows = int(number), "Csv should have {number} rows but has {number}"
```

features/steps/data.py
```diff
@@ -46,6 +48,48 @@ def step_impl(context):
...
+
+
+@then("the csv output has the columns")
+def step_impl(context):
+
+    for column in context.table:
+        assert column in context.csv_output.columns
```
mikeAdamss: include a message, see other comment

features/steps/data.py
```diff
@@ -46,6 +48,48 @@ def step_impl(context):
...
+@then("I read the metadata output '{metadata_output}'")
+def step_impl(context, metadata_output):
+    try:
+        json_output_path = Path(metadata_output)
+
+        metadata_file = open(json_output_path)
+
+        context.json_output = json.load(metadata_file)
+
+    except Exception as exc:
```
mikeAdamss: no need to catch and raise, just let it raise, see other comment

features/steps/data.py
```diff
@@ -46,6 +48,48 @@ def step_impl(context):
...
+        raise exc
+
+
+@then("the metadata should match 'cpih-metadata-correct.json'")
+def step_impl(context):
+    expected_metadata = {"to": "do"}
+
+    assert context.json_output == expected_metadata
```
mikeAdamss: include a message, see other comment

features/environment.py
```diff
@@ -22,7 +22,10 @@ def before_all(context):
     Also checks if the fixtures unzipped data path exists,
     and if it doesn't, extracts the fixtures zipfile to it.
     """
-    context.fixture_destination_path = Path("features/fixtures/data")
```
mikeAdamss, line 25: now we've got the ability to, can you add in a few lines to use monkeypatch to set `DISABLE_NOTIFICATIONS` to true please

features/data_ingress_v1.feature
```diff
@@ -11,14 +11,20 @@ Feature: Data Ingress v1
...
         | file                   |  fixture                           |
         | pipeline-config.json   |  pipeline_config_basic_valid.json  |
-        | data.xml               |  data_sdmx_valid_1.sdmx            |
+        | data.xml               |  data.xml                          |
     And v1_data_ingress starts using the temporary source directory
     Then the pipeline should generate no errors
+    And I read the csv output 'data.csv'
+    And the csv output should have '9744' rows
+    And the csv output has the columns
+          | Column 1 | Column 2 | Column 3 |
```
mikeAdamss: something isn't right here, the expected behaviour of this line (in. this case) is enforcing that the csv has 3 columns which are named "Column 1", "Column 2" and "Column 3". that's not the data we have.

features/data_ingress_v1.feature
```diff
@@ -11,14 +11,20 @@ Feature: Data Ingress v1
     Given a temporary source directory of files
         | file                   |  fixture                           |
         | pipeline-config.json   |  pipeline_config_basic_valid.json  |
-        | data.xml               |  data_sdmx_valid_1.sdmx            |
+        | data.xml               |  data.xml                          |
```
mikeAdamss: you probably do need a more verbose/descriptive name, remember that over time there'll be many different valid data fixtures in the zip file. I'd find out what dataset it is (cpih? unsure) and include that in the zipped name somehow.

features/data_ingress_v1.feature
```diff
@@ -11,14 +11,20 @@ Feature: Data Ingress v1
...
+          | Column 1 | Column 2 | Column 3 |
+    And I read the metadata output 'metadata.json'
+    And the metadata should match 'cpih-metadata-correct.json'
 
   Scenario: Pipeline runs with an expected error
     Given a temporary source directory of files
         | file                            |  fixture                           |
         | incorrectly-named-config.json   |  pipeline_config_basic_valid.json  |
-        | data.xml                        |  data_sdmx_valid_1.sdmx            |
+        | data.xml                        |  data.xml                          |
```
mikeAdamss: see other comment on naming

mikeAdamss (commented):
a few comments but good job on this, not a straight forward task

features/data_ingress_v1.feature
NickPapONS: Resolved but note: I made it check for the existence of some columns from the output, I didn't make it check all the columns because there is a very large number of them and I didn't want to put so many of them in the feature file table.

features/data_ingress_v1.feature
mikeAdamss: you cant call it data Nick, there will be other xml files with data in that zip , needs to be more descriptive.

features/environment.py
```diff
@@ -8,6 +8,7 @@
 import requests
 import docker
 from docker import DockerClient
+from pytest import MonkeyPatch
```
mikeAdamss: bad call by me, see main pr comment.

features/steps/data.py
```diff
@@ -46,6 +48,40 @@ def step_impl(context):
...
+    json_output_path = Path(metadata_output)
+
+    metadata_file = open(json_output_path)
+
+    context.json_output = json.load(metadata_file)
+
+
+@then("the metadata should match '{correct_metadata}'")
+def step_impl(context, correct_metadata):
+    expected_metadata = {"to": "do"}
```
mikeAdamss: corrected metadata is a path, the data at that path should be loaded to become expected_metadata - that's what you need to compare to context.json.output

mikeAdamss (commented):
couple of minor comments where you've marked something as resolved but I think missed the commit.

mainly though there's a nasty bit here to sort out around test setup.

## thing 1

there are two problems in the pipeline that have been fixed elsewhere, these two lines:

- https://github.com/ONSdigital/dp-data-pipelines/blob/11f5f41dee98994c4ec2438f1c45a0d3db318713/dpypelines/s3_tar_received.py#L13
-https://github.com/ONSdigital/dp-data-pipelines/blob/11f5f41dee98994c4ec2438f1c45a0d3db318713/dpypelines/pipeline/dataset_ingress_v1.py#L15

are running on import, which was a bad idea in hindsight, they both need to be moved down to be the first things that happens in the function below them.

that will make this easier to control as we can set the env var before we call the function while testing.


## thing 2

I was wrong about monkyepatch, its a pytest module (doh), it only works with pytest 😩 , so we'll have to do this the old fashioned way.

something like:

```python
def before_all(context):
     
     context.original_notification_setting = os.environ.get("DISABLE_NOTIFICATIONS", None)
     os.envion["DISABLE_NOTIFICATIONS"] = "true"
```

then 

```
def after_all(context):

     if context.disable_notification_setting is not None:
           os.environ["DISABLE_NOTIFICATIONS"] = context.disable_notification_setting
```

All taken together. it'll let is toggle off notifications for the length of the tests but always put the users environment back how it was after the tests have ran.

sorting that bit out aside. this is very close 😄

features/data_ingress_v1.feature
NickPapONS: Yep I forgot this one

SarahJohnsonONS (approved):
LGTM - confirmed that `DISABLE_NOTIFICATIONS` is being changed/changed back as appropriate in `environment.py`

PR #96, opened 2024-04-08
SDMX Transform - Sanity check SDMX file (validity)

bailed22:


dpypelines/pipeline/shared/transforms/sanity_check.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+def sdmx_sanity_check_v1(sdmx_file: Path):
+    """
+    Sanity check that the recieved sdmx file is actually an sdmx file.
+    """
+    assert sdmx_file.name.endswith(".xml"), "Invalid sdmx_file"
+
+    try:
+        pd.read_xml(sdmx_file, nrows=10)
```
JimBryant7: The function fails for me. From my end 'nrows' is not a valid kwarg for read_xml(), does it work for you or is it a python version issue?

bailed22: Have restructure the script and run correctly now

JimBryant7 (commented):
Can you just rename the test file to test_sanity_check so that it matches the sanity_check file please

JimBryant7 (approved):

PR #97, opened 2024-04-09
add configurable notification postfix

mikeAdamss:
### What

Added the ability to add a string postfix to the end of the success and fail notifications (so we can add a url at the platform level).

Have removed the enrich_online decorator, made sense when we were sending full messages, less so now.

### How to review

Sanity check,

### Who can review

Anyone.


Moasib-Arif (approved):
The Code seems to LGTM 👍

PR #98, opened 2024-04-09
Optimize sdmx transform script 

osamede20:
### What
1. The codes were refactored. The process of getting the Header dictionary remains the same. The Python `ElementTree` package could not be used for the header dictionary as most of the header elements do not have attributes.
2. Two new functions were created - `flatten_dict` and `convert`. The `flatten_dict` flattens the nested Header dictionary and the `convert` converts the tuple dictionary to a flat 'concatenated' dictionary. The two functions are included in the utils.py
3. The Python `ElementTree` package was used to retrieve the observation block dictionaries and the parent series block dictionaries with which each observation block shares common headers.  

### How to review
The codes can be checked for performance optimisation to see if the processing time could be further reduced from about 10s.
They can also be checked for scalability. 

### Who can review
Any DE can review.

https://jira.ons.gov.uk/browse/DIS-472

### DIFF
--- a/dpypelines/pipeline/shared/transforms/sdmx/compact/v20/prototype/v1.py
+++ b/dpypelines/pipeline/shared/transforms/sdmx/compact/v20/prototype/v1.py
@@ -51,6 +51,11 @@ def xmlToCsvSDMX2_0(input_path, output_path):
     )
 
     full_table = pd.concat([header_frame, series_frame, obs_frame], axis=1)
+
+    # the following is just tidying up the column headers so they are not filled with @ and such
+    header_replace = { x: str(x).replace('@', '') for x in full_table.columns }
+    full_table.rename(columns=header_replace, inplace=True)
+
     full_table.to_csv(output_path, encoding="utf-8", index=False)
     return full_table
 
--- a/dpypelines/pipeline/shared/transforms/sdmx/compact/v20/prototype/v1.py
+++ b/dpypelines/pipeline/shared/transforms/sdmx/compact/v20/prototype/v1.py
@@ -21,7 +21,7 @@ def xmlToCsvSDMX2_0(input_path, output_path):
 
     # Attributes are name–value pair that exist within a start-tag or empty-element tag. An XML attribute can only have a single value and each attribute can appear at most once on each element.
 
-    # The element tag is retrieved as .tab on Python ElementTree package. 
+    # The element tag is retrieved as .tab on Python ElementTree package.
     # The name-value pair attributes are retrieved as .attrib on Python ElementTree package.
 
     # Each block of observations tagged "Obs" within the same column headers is contained within a parent block tagged "Series". The respective tuples of dictionary are converted into 'flat' dictionaries and dropped into the series_dict and obs_dict.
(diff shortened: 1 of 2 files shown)


dpypelines/pipeline/shared/transforms/sdmx/compact/v20/prototype/v1.py
```diff
@@ -1,99 +1,68 @@
...
-                    for obs in entry.items():
-                        temp = []
-                        temp.append(obs[1])
-                        temp_df[obs[0]] = temp
-                    obs_df = pd.concat([obs_df,temp_df])
-            
-        table_df = pd.concat([headers_df, obs_df], axis = 1) # merge the headers to each observation
+    # Here the records on the header dataframe are replicated to match the length of the series and observation dataframes.
+    repl_rows = header_df.loc[0].copy()
+    header_frame = header_df._append(
```
JimBryant7: This failed for me when I ran it - _append wasn't recognised
I changed it from header_df._append to header_df.append and it worked
Is this just a typo or is it something I'm missing on my end?

dpypelines/pipeline/shared/transforms/sdmx/compact/v20/prototype/v1.py
```diff
@@ -1,99 +1,68 @@
...
+        [repl_rows] * (len(obs_frame) - 1), ignore_index=True
+    )
 
-        output.append(table_df) # add to a list of all the series blocks that have been processed 
-
-    full_table = pd.concat(output) # merge all the series blocks together
-
-    # the following is just tidying up the column headers so theyre not filled with @ and such
-
-    header_replace = { x: str(x).replace('@', '') for x in full_table.columns }
```
JimBryant7: I think these 2 rows should be left in for consistency (removing the @'s from the column names)

dpypelines/pipeline/shared/transforms/sdmx/compact/v20/prototype/v1.py
osamede20: That's interesting!`.append` does not work with mine; `._append `works for me.
I believe this is something to investigate or it might just have to do with our python interpreter.
I am using Python 3.9.16 (dpypelines-py.3.9)

I am happy to change it to `.append` if the issue is from my python package.

dpypelines/pipeline/shared/transforms/sdmx/compact/v20/prototype/v1.py
osamede20: Absolutely! I missed the two lines of codes.
I have included them.

dpypelines/pipeline/shared/transforms/sdmx/compact/v20/prototype/v1.py
JimBryant7: I've found this from stack overflow https://stackoverflow.com/questions/75956209/error-dataframe-object-has-no-attribute-append
apparently append has been removed from pandas (I'm using python 3.7 so its still there and you are using 3.9 so that is why it is failing for you), _append is "the method is private and is not intended to be used outside of pandas' internal code"
the suggestion is to use concat instead. So I think it should be changed to concat if possible so theres not issues

osamede20: Sorted; thanks.

JimBryant7 (approved):

PR #99, opened 2024-04-09
Configuration dictionary added

SarahJohnsonONS:
### What

The `pipeline-config.json` is being replaced with a dictionary (`CONFIGURATION` in `dpypelines/pipeline/configuration.py`) containing the details required to run a transform on a specific dataset. These details will be accessed via a `dataset_id`, which is used as the `CONFIGURATION` dictionary key (actually a regex pattern that matches the `dataset_id`).

The `dataset_id` should be specified as part of the input metadata. This value can then be passed to the relevant `dataset_ingress` function (`CONFIGURATION[secondary_function]`), with the correct configuration details.

```python
from dpypelines.pipeline.dataset_ingress_v1 import dataset_ingress_v1
from dpypelines.pipeline.shared.transforms.sdmx.v1 import (
    sdmx_sanity_check_v1,
    sdmx_compact_2_0_prototype_1,
)

# Regex pattern matching `dataset_id` as dictionary key
CONFIGURATION = {
    # Default configuration (regex pattern matches any string of characters of length >= 0)
    "^.*$": {
        "config_version": 1,
        "transform": sdmx_compact_2_0_prototype_1,
        "transform_kwargs": {},
        "required_files": ["^data.xml$"],
        "supplementary_distributions": ["^data.xml$"],
    },
    # `cpih` config details
    "^cpih$": {
        "config_version": 1,
        "transform": sdmx_compact_2_0_prototype_1,
        "transform_kwargs": {},
        "required_files": ["^data.xml$"],
        "supplementary_distributions": ["^data.xml$"],
    },
}
```

### How to review

Make sure you understand how the values in `CONFIGURATION` will be used to configure a pipeline, and that all fields required to run `dataset_ingress_v1()` are present.

### Who can review

Anyone.


dpypelines/pipeline/configuration.py
```diff
@@ -0,0 +1,30 @@
...
+from dpypelines.pipeline.shared.transforms.sdmx.v1 import (
+    sdmx_compact_2_0_prototype_1,
+    sdmx_sanity_check_v1,
+)
+
+# Set a regex pattern matching the `dataset_id` as `CONFIGURATION` dictionary key
+# All fields are required in order for a pipeline transform to run successfully
+CONFIGURATION = {
+    # Default configuration (regex pattern matches any string of characters of length >= 0)
+    "^.*$": {
```
mikeAdamss: so with matching you have to be careful of order, if you think of someone using this they'd check for matches in order (iterate), you'd need to do it that way as more that one pattern will/could/should match - so you're looking for the _best_ match.

so what we want here would be:

- check first key - if cpih matches we use that dict
- if its not check the next option -  with the last option being the default as that'll match any idenfifer.

as-in its impossible to match "cpih" here as the first key will always match

dpypelines/pipeline/configuration.py
```diff
@@ -0,0 +1,30 @@
...
+        "config_version": 1,
+        "transform": sdmx_compact_2_0_prototype_1,
+        "transform_inputs": {"^data.xml$": sdmx_sanity_check_v1},
+        "transform_kwargs": {},
+        "required_files": ["^data.xml$"],
+        "supplementary_distributions": ["^data.xml$"],
+        "secondary_function": dataset_ingress_v1,
+    },
+    # `cpih` config details
```
mikeAdamss: after you move this, can you just a comment explaining "this is an example of how you set config for a specific dataset idnetifier" (or similar). Just so the next person looking at this knows why we've defined the same config twice

dpypelines/pipeline/configuration.py
mikeAdamss: side note: there's a common misconception that dictionaries are not ordered. That _used_ to be the case (<python 3.7) as python would finagle the order of things to make it more memory efficient, they solved that a while ago though on account of it being silly.

mikeAdamss (commented):
minor comments

dpypelines/pipeline/configuration.py
SarahJohnsonONS: thanks, i'd been wondering about that XD

mikeAdamss (approved):
LGTM 👍

Issue #90, opened 2024-03-27
data_ingress_v1 updated for new configuration


mikeAdamss:
DEPENDS ON: https://github.com/ONSdigital/dp-data-pipelines/issues/88

## What is this

Once we have an updated appraoch to configuration we need to update https://github.com/ONSdigital/dp-data-pipelines/blob/sandbox/dpypelines/pipeline/dataset_ingress_v1.py to use it.

## What to do

new function signature

```python
def dataset_ingresss_v1(path_to_files, config):
```

- The signature needs to expect two things
    - a config entry from the DEPENDS ON task
    - the path to the local files (as it currently does)
- From that starting point rewrite the function, it should mainly be simplification (as we dont need to check for config anymore)
- You'll also need to change the logic for getting required files and supplemntary distributions, there's a "config_version" key in the DEPENDS_ON task that should serve the same purpose and allow us to change the
structure of config dicts in the future should we want to.

## Acceptance Criteria

- [ ] The function will run using the new config approach.

mikeAdamss:
done

Issue #89, opened 2024-03-27
update s3_tar_recieved for new config approach


mikeAdamss:
DEPENDS ON: https://github.com/ONSdigital/dp-data-pipelines/issues/88

## What is this

Once we have an updated appraoch to configuration we need to update https://github.com/ONSdigital/dp-data-pipelines/blob/50846d7d1b8c14a4cced4349144d4b83066bdd02/dpypelines/s3_tar_received.py#L14 to use it.

## What to do

- Write a function to use the recieved s3 url to get the correct entry from our new config (see DEPENDS ON task).
- Remove all the pipeline config stuff
- Get the "secondary_function" and call it where we're calling `dataset_ingress_v1`

Note: you'll need to pass along two positonal arguments to the function at the end. `<config>, <path_to_inputs>`  

## Acceptance Criteria

- [ ] Pipeline will run as far as calling the "secondary_function" with the path to the decompressed files as the argument.

SarahJohnsonONS:
can you set the first example "key" to "^.*$" (anything - so it'd set our wanted default of calling that sdmx transform)
if you do pick up the s3 tar received task create an empty function for get_dataset_id or similar that just returns a hard coded "not-specified" (or suchlike for now),.
function to get the config based on the key - your "not-specified" will match the default
only get the config once, just pass the config dictionary on to the next function

mikeAdamss:
done

PR #101, opened 2024-04-10
#89 update s3 tar recieved for new config approach

SarahJohnsonONS:
### What

#89:
- Replaced existing pipeline config sections of `s3_tar_received.start()`  with `CONFIGURATION` dictionary approach.
- `get_dataset_id()` placeholder function added, to be updated once we know where the `dataset_id` can be extracted from.
- `get_pipeline_config()` function added to get the correct configuration details from `CONFIGURATION` (with unit tests).
- `start()` now calls the function specified in the `CONFIGURATION["secondary_function"]` field, rather than calling `dataset_ingress_v1()` directly. 

#90:
- `pipeline_config` added to `dataset_ingress_v1()` arguments.
- Replaced existing pipeline config sections of `dataset_ingress_v1()` with `CONFIGURATION` dictionary approach.
- `get_required_files_patterns()` and `get_supplementary_distributions_patterns()` updated.
- `dpypelines/pipeline/shared/config.py` and `dpypelines/pipeline/shared/details.py` deleted as no longer needed, along with associated tests, fixtures and test cases.
- Acceptance tests amended - new steps for valid/invalid pipeline config, `config.json` removed from `data-fixtures.zip` and feature file tables.

### How to review

Sanity check, make sure it makes sense. I was getting a credentials error when trying to run `start()` with an actual S3 bucket, so if you can get it to run and confirm it works, that would be great. One question - previously, we were using the `local_store` to gain access to the `has_lone_file_matching()` and `get_lone_matching_json_as_dict()` methods. Now it only uses the `get_current_source_pathlike()` method, but the local store is created again in the `dataset_ingress_v1()` function - it might be possible to tidy this up a bit - suggestions welcome.

### Who can review

Anyone.


dpypelines/s3_tar_received.py
```diff
@@ -46,56 +48,29 @@ def start(s3_object_name: str):
...
-        if not local_store.has_lone_file_matching(r"^pipeline-config.json$"):
-            de_notifier.failure()
-            msg = message.expected_local_file_missing(
-                "Pipeline config not found",
-                "pipeline-config.json",
-                "dataset_ingress_v1",
-                local_store,
-            )
-            raise ValueError(msg)
+        for key in CONFIGURATION.keys():
```
mikeAdamss: you'll need a break statement here, currently even if it match the first key it'll keep going and match the default as well - which will be the one that plays as it'll be the last match.

I'd go one further tbh and make a utility `get_config()` function or similar. Regex can be faffy so a function (with accompanying unit tests to make sure it's doing what we think) is probably wise.

pyproject.toml
```diff
@@ -11,7 +11,6 @@ python = ">=3.9.0,<3.10.0"
 dpytools = { git = "https://github.com/ONSdigital/dp-python-tools.git", branch="develop" }
 pandas = "^2.2.1"
 unidecode = "^1.3.8"
-xmltodict = "^0.13.0"
```
mikeAdamss: I think the DEs use this: https://github.com/ONSdigital/dp-data-pipelines/blob/f36ea456a4490730399d69ef34878d7cf10d0030/dpypelines/pipeline/shared/transforms/sdmx/compact/v20/prototype/v1.py#L5

dpypelines/s3_tar_received.py
```diff
@@ -46,56 +48,29 @@ def start(s3_object_name: str):
             )
         ) from err
 
-    # Check for the existence of a configuration file
+    # get_dataset_id() currently empty - returns "not-specified"
+    # To be updted once we know where the dataset_id can be extracted from
```
mikeAdamss: can you add that working from the s3 object name is an assumption, might be we do, might not be, tbd.

mikeAdamss (commented):
loving how much code we're getting to remove 😄 

only one not faffy comment around pulling out the matching logic, I don't trust regex and want it nailed down.

we'll also need to update the feature logic. (the bit where we provide a pipeline config) - for that can you see what'd need to change and if it's trivial (I suspect so) then make the change but clue nick into it, he's changing a lot in there at the moment.

you've got a failed quality check btw, isort not ran I think. Not you but you'll need to do it.

dpypelines/s3_tar_received.py
```diff
@@ -1,16 +1,16 @@
 from dpytools.s3.basic import decompress_s3_tar
 from dpytools.stores.directory.local import LocalDirectoryStore
-from dpytools.validation.json import validation
 
-from dpypelines.pipeline.dataset_ingress_v1 import dataset_ingress_v1
+from dpypelines.pipeline.configuration import (
+    CONFIGURATION,
```
mikeAdamss: you don't need `CONFIGURATION`, see other comment

dpypelines/s3_tar_received.py
```diff
@@ -46,56 +50,27 @@ def start(s3_object_name: str):
...
-        if not local_store.has_lone_file_matching(r"^pipeline-config.json$"):
-            de_notifier.failure()
-            msg = message.expected_local_file_missing(
-                "Pipeline config not found",
-                "pipeline-config.json",
-                "dataset_ingress_v1",
-                local_store,
-            )
-            raise ValueError(msg)
+        pipeline_config = get_pipeline_config(dataset_id, CONFIGURATION)
```
mikeAdamss: now we're moved the function the `CONFIGURATION` constant is in the same module as the function, no need to pass it around.

mikeAdamss: see other follow up comment on the. import

dpypelines/s3_tar_received.py
mikeAdamss: been thinking about this, two options that I can see:

1.) we yes import it, but set it via a keyword argument on `start()`, that way when we come to acceptance tests for s3_tar_received we it'll make our lives easy.

2.) we don't import it (as `get_pipeline_config()` already knows what `CONFIGURATION` is, its in the same file. and we'll work out some way of setting it for tests when we get to it.

I'm happy with either but right now we're importing CONFIGURATION and passing it to a file which already knows that value, doesn't make sense.

dpypelines/s3_tar_received.py
```diff
@@ -46,56 +50,24 @@ def start(s3_object_name: str):
...
     except Exception as err:
         de_notifier.failure()
         raise Exception(
             message.unexpected_error(
-                "Error while checking for pipeline-config.json", err
+                "Failed to get pipeline configuration details", err
```
mikeAdamss: I missed this previously but can you add some details to this error. so what the identifier was and what the keys in CONFIGURATION were, something like:

```
Failed to match dataset id "foo" to a configuration from ["bar", "baz"]
```
enough we can see what didnt match without digging.

features/steps/data.py
```diff
@@ -1,6 +1,32 @@
...
 from pathlib import Path
-import os
+
+from dpypelines.pipeline.dataset_ingress_v1 import dataset_ingress_v1
+from dpypelines.pipeline.shared.transforms.sdmx.v1 import (
+    sdmx_compact_2_0_prototype_1,
+    sdmx_sanity_check_v1,
+)
+
+configuration = {
```
mikeAdamss: `CONFIGURATION`

features/data_ingress_v1.feature
```diff
@@ -9,8 +9,8 @@ Feature: Data Ingress v1
   Scenario: Pipeline runs without errors
     Given a temporary source directory of files
         | file                   |  fixture                           |
-        | pipeline-config.json   |  pipeline_config_basic_valid.json  |
         | data.xml               |  esa2010_test_data.xml             |
+    And a pipeline configuration dictionary that is 'valid'
```
mikeAdamss: this is probably `Given a dataset id of "valid"`.

mikeAdamss (commented):
well done, this looks good to me 😄 , minor pedantry only.

one extra thing. - can you add a new github actions to run features (make it a separate action please). Now is working lets keep it that way.

dpypelines/s3_tar_received.py
SarahJohnsonONS: I've done this, but because I've removed the `CONFIGURATION` import, I had to add a new `config_keys` return value to `get_pipeline_config` so that the available keys can be included in the error message. If there's a better way, let me know.

dpypelines/s3_tar_received.py
```diff
@@ -46,56 +46,25 @@ def start(s3_object_name: str):
             )
         ) from err
 
-    # Check for the existence of a configuration file
+    # get_dataset_id() currently empty - returns "not-specified"
+    # To be updated once we know where the dataset_id can be extracted from (not necessarily s3_object_name as suggested by argument name)
+    dataset_id = get_dataset_id(s3_object_name)
```
mikeAdamss: needs a catch, if this fails for whatever reason we'll need to be notified.

mikeAdamss: I know it cant fail _now_ btw, but lets be consistent, all processes should have a catch etc.

mikeAdamss (commented):
this looks good 👍 , well done.

one catch to add and happy to approve.

dpypelines/s3_tar_received.py
mikeAdamss: fair point, that'll do fine for now

mikeAdamss (approved):
LGTM

PR #115, opened 2024-04-22
Added s3_tar_recieved_documentation

Moasib-Arif:
### What

Added information to the readme for the repo on how to run s3_tar_received.start()

### How to review

Sanity check,  Check for any typos or try runing the s3_tar_recieved.start() funtion using the information provided in the readme

### Who can review

Anyone

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,11 +15,14 @@ This repository is provided as an installable python package
     pip install git+https://github.com/ONSdigital/dp-data-pipelines.git
     ```
 
+## Setup
+
+Before you start using the `dp-data-pipelines` , you need to set up your environment:
+
+1. **AWS SSO Login**: Ensure you are logged in to AWS via SSO. This is necessary for the function to access the S3 bucket and process the tar file. If you haven't logged in, you can do so by running `aws sso login` in your terminal.
 
 ## Usage
 This package provides multiple data pipelines. The initial use case is the s3_tar_received.start() function, which handles the required behavior when receiving a tar file indicated by an S3 object name.
-
-### Before using the `s3_tar_received.start()` function, ensure you are logged in to AWS via SSO. This is necessary for the function to access the S3 bucket and process the tar file.
 
 Here's a basic example of how to use it:
 
@@ -42,6 +45,18 @@ The `s3_tar_received.start()` function performs the following steps:
 If the `s3_tar_received.start()` function encounters any problems, an error will be raised that includes some information on the issue encountered, to help you resolve the problem. For example:
 
 -   If it is not possible to create the local store from the decompressed tar file, an error message will be displayed indicating the issue with creating the local directory store.
+
+
+## Configuration
+
+| Environment variable  | Default | Description                                                               |
+|:----------------------|:-------:|:--------------------------------------------------------------------------|
+| DE_SLACK_WEBHOOK      |  None   | Webhook for the Data Engineering Slack channel                            |
+| SE_SLACK_WEBHOOK      |  None   | Webhook for the Software Engineering Slack channel                        |
+| DS_SLACK_WEBHOOK      |  None   | Webhook for the Data Submitters Slack channel                             |
+| PS_SLACK_WEBHOOK      |  None   | Webhook for the Publishing Support Slack channel                          |
+| DISABLE_NOTIFICATIONS |  False  | Toggle Slack notifications on or off                                      |
+| AWS_PROFILE           |  None   | The AWS environment (dp-bleed-dev, dp-staging, dp-sandbox, dp-production) |
 
 Licence
 -------
--- a/README.md
+++ b/README.md
@@ -27,11 +27,9 @@ s3_tar_received.start('my-bucket/my-data.tar')
 ```
 In this example, 'my-bucket/my-data.tar' is the S3 object name of the tar file to be processed.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -16,8 +16,33 @@ This repository is provided as an installable python package
...
+
+s3_tar_received.start('my-bucket/my-data.tar')
+```
+In this example, 'my-bucket/my-data.tar' is the S3 object name of the tar file to be processed.
+
+The tar file should include a pipeline-config.json. The configuration for the pipeline is defined in the configuration.py file. This JSON file contains the configuration details for the pipeline, such as the dataset id, the transform function to be used, the required files, and the function to be called after the main processing function (referred to as the "secondary function").
+
+The CONFIGURATION dictionary in this file contains the configuration details for each dataset. The get_dataset_id() function is used to extract the dataset id from the S3 object name, and the get_pipeline_config() function is used to get the pipeline configuration details for the given dataset id.
+
+For more information about the pipeline-config.json file, refer to the schema [here](dpypelines/schemas/dataset-ingress/config/v1.json).
```
SarahJohnsonONS: Remove this, as we aren't using the config.json approach, so no schema needed.

README.md
```diff
@@ -16,8 +16,33 @@ This repository is provided as an installable python package
...
+Here's a basic example of how to use it:
+
+```python
+from dpypelines import s3_tar_received
+
+s3_tar_received.start('my-bucket/my-data.tar')
+```
+In this example, 'my-bucket/my-data.tar' is the S3 object name of the tar file to be processed.
+
+The tar file should include a pipeline-config.json. The configuration for the pipeline is defined in the configuration.py file. This JSON file contains the configuration details for the pipeline, such as the dataset id, the transform function to be used, the required files, and the function to be called after the main processing function (referred to as the "secondary function").
```
SarahJohnsonONS: The tar file no longer needs to include a pipeline-config.json file - all config details are accessed from the `CONFIGURATION` dictionary in `configuration.py`. At the moment, the `get_dataset_id` function returns "not-specified", which in turn is passed to `get_pipeline_config`, which returns the default configuration details (as "not-specified" matches the "^.*$" key). We don't know how the `dataset_id` will be extracted from the tar file yet - it'll *probably* be from the S3 object name, but might be best not to include this in the documentation, as it may need to be changed.

README.md
```diff
@@ -16,8 +16,33 @@ This repository is provided as an installable python package
...
+For more information about the pipeline-config.json file, refer to the schema [here](dpypelines/schemas/dataset-ingress/config/v1.json).
+
+```python
+s3_tar_received.start()
+```
+The s3_tar_received.start() function performs the following steps:
+
+1. Decompresses the tar file to the workspace.
+2. Creates a local directory store using the decompressed files.
+3. Extracts the dataset ID from the S3 object name.
```
SarahJohnsonONS: See comment above - this *might* be the way we get the dataset ID, but it's not finalised yet. Probably best just to remove this line for now.

README.md
```diff
@@ -16,8 +16,33 @@ This repository is provided as an installable python package
...
+4. Retrieves the configuration details for the given dataset ID from the pipeline-config.json file.
```
SarahJohnsonONS: Change "dataset ID" to just "dataset".
Change "from the pipeline-config.json file" with "from `configuration.py`".

README.md
```diff
@@ -16,8 +16,33 @@ This repository is provided as an installable python package
...
+5. Calls the secondary function specified in the pipeline configuration. This function is defined in the dataset_ingress_v1.py file and is responsible for ingesting the data into the system.
```
SarahJohnsonONS: This needs to be a bit more general - rather than referring to `dataset_ingress_v1`, say something like "The secondary function defines which version of the transform functionality will be applied to the dataset."

SarahJohnsonONS (commented):
Might be good to have some info on errors - e.g. "if the `s3_tar_received.start()` function encounters any problems, an error will be raised that includes some information on the issue encountered, to help you resolve the problem. For example, if it is not possible to create the local store from the decompressed tar file... <insert example of the error that would be raised>"

README.md
```diff
@@ -16,8 +16,32 @@ This repository is provided as an installable python package
     ```
 
 ## Usage
```
SarahJohnsonONS, line 22: Add something here about needing to log in to AWS via SSO for `start()` to work.

README.md
```diff
@@ -16,8 +16,32 @@ This repository is provided as an installable python package
...
+```python
+from dpypelines import s3_tar_received
+
+s3_tar_received.start('my-bucket/my-data.tar')
+```
+In this example, 'my-bucket/my-data.tar' is the S3 object name of the tar file to be processed.
+
+The CONFIGURATION dictionary in this file contains the configuration details for each dataset and the function to be called after the main processing function (referred to as the "secondary function"). The get_dataset_id() function is used to extract the dataset id, and the get_pipeline_config() function is used to get the pipeline configuration details for the given dataset id.
+
+```python
```
SarahJohnsonONS: I think this code block can be deleted.

SarahJohnsonONS (commented):
Almost there, could you format any text that refers to modules/variables/functions etc as inline code blocks (i.e. `s3_tar_received.start()` instead of s3_tar_received.start(), `CONFIGURATION` instead of CONFIGURATION, `configuration.py` instead of configuration.py). Makes it clearer to end users which bits are references to specific code.

SarahJohnsonONS (approved):
LGTM

Issue #109, opened 2024-04-17
logging - data ingress v1


mikeAdamss:
## What is this

Add logging statments to data_ingress_v1.

## What to do

These are the docs for the logger:
https://github.com/ONSdigital/dp-python-tools/tree/develop/dpytools/logging

And I've added some quick examples of using it here:
https://github.com/ONSdigital/dp-data-pipelines/blob/logging-trial/dpypelines/s3_tar_received.py

There's no "bad" place to add logs but its hekpful to have a consistent pattern, so for this task you're looking to log out:

- when we've successfully done a thing (with appropriate variables)
- when we've tried to do a thing and got an error (with appropriate variables)

This **does** change the approach in the try catches, so this:

```python
try:
    decompress_s3_tar(s3_object_name, "input")
    except Exception as err:
        de_notifier.failure()
        raise Exception(
            message.unexpected_error(
                f"Failed to decompress tar file {s3_object_name}", err
            )
        ) from err
```

becaomes something like this

```python
try:
    decompress_s3_tar(s3_object_name, "input")
    logger.info("S3 tar recieved successfully decompressed to ./input", data={"s3_object_name": s3_object_name})
except Exception as err:
    logger.error("Failed to decompress tar file", err, data={"s3_object_name": s3_object_name})
    de_notifier.failure()
    raise err
```

the overarching point being we dont need the verbose expression of what went wrong as the logger will capture
and format that information for us.

note: no tests needed for this, the logger has tests already.

## Acceptance criteria

- [ ] s3 tar recieved has logging

PR #116, opened 2024-04-22
#109 logging v1 data ingress

NickPapONS:
### What

Adds logging to v1_data_ingress. Logging was mainly added to any case where something is being tried, a logger statement at info level for if it succeeds and a logger error statement if an error occurs.

### How to review

Look through the instances where the logger is being used and see if the message makes sense. Also check if the `data` argument is being given relevant information/variables to that log, or if I could add another variable to make that log more informative.

### Who can review

Anyone

### DIFF
--- a/dpypelines/pipeline/dataset_ingress_v1.py
+++ b/dpypelines/pipeline/dataset_ingress_v1.py
@@ -100,7 +100,7 @@ def dataset_ingress_v1(files_dir: str, pipeline_config: dict):
     for match, sanity_checker in pipeline_config["transform_inputs"].items():
         try:
             input_file_path: Path = local_store.save_lone_file_matching(match)
-            logger.info(f"Successfully saved file that matches pattern", data={"input_file_path": input_file_path, "match": match})
+            logger.info("Successfully saved file that matches pattern", data={"input_file_path": input_file_path, "match": match})
         except Exception as err:
             files_in_directory = local_store.get_file_names()
             logger.error("Error occured while attempting to save matching pattern file.", 
@@ -145,7 +145,6 @@ def dataset_ingress_v1(files_dir: str, pipeline_config: dict):
         err, 
         data={"transform_function": transform_function, 
         "pipeline_config": pipeline_config,
-        "transform_function": transform_function,
         "args": args,
         "kwrags": kwargs
         })
--- a/dpypelines/pipeline/dataset_ingress_v1.py
+++ b/dpypelines/pipeline/dataset_ingress_v1.py
@@ -1,7 +1,9 @@
 import json
+from os import pipe
 from pathlib import Path
 
 from dpytools.stores.directory.local import LocalDirectoryStore
+from dpytools.logging.logger import DpLogger
 
 from dpypelines.pipeline.shared import message
 from dpypelines.pipeline.shared.notification import (
@@ -10,6 +12,7 @@ from dpypelines.pipeline.shared.notification import (
 )
 from dpypelines.pipeline.shared.pipelineconfig import matching
 
+logger = DpLogger('data-ingress-pipeline')
 
 def dataset_ingress_v1(files_dir: str, pipeline_config: dict):
     """
@@ -34,25 +37,28 @@ def dataset_ingress_v1(files_dir: str, pipeline_config: dict):
     # Attempt to access the local data store
     try:
         local_store = LocalDirectoryStore(files_dir)
+        logger.info(f"Local data store successfully instansiated", data ={"local_store_dir": files_dir, "local_store": local_store})
     except Exception as err:
+        logger.error(f"Failed to access local data at {files_dir}", err, data={"local_store_dir": files_dir})
         de_notifier.failure()
-        raise Exception(
-            message.unexpected_error(f"Failed to access local data at {files_dir}", err)
-        ) from err
+
+        raise err
 
     # Extract the patterns for required files from the pipeline configuration
     try:
         required_file_patterns = matching.get_required_files_patterns(pipeline_config)
+        logger.info(f"Required file patterns retrieved from pipeline configuration", data={"required_file_patterns": required_file_patterns})
     except Exception as err:
+        logger.error("Failed to get required files pattern", err, data={"pipeline_config": pipeline_config, "local_store": local_store})
         de_notifier.failure()
-        raise Exception(
-            message.unexpected_error("Failed to get required files patterns", err)
-        ) from err
+
+        raise err
 
     # Check for the existence of each required file
     for required_file in required_file_patterns:
         try:
             if not local_store.has_lone_file_matching(required_file):
+                logger.info(f"Required matching file {required_file} was not found.", data={"required_file": required_file})
                 de_notifier.failure()
                 msg = message.expected_local_file_missing(
                     f"Required file {required_file} not found",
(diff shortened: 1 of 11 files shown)


dpypelines/pipeline/dataset_ingress_v1.py
```diff
@@ -34,25 +36,28 @@ def dataset_ingress_v1(files_dir: str, pipeline_config: dict):
...
     try:
         local_store = LocalDirectoryStore(files_dir)
+        logger.info("Local data store successfully instansiated", data ={"local_store_dir": files_dir, "local_store": local_store})
     except Exception as err:
+        logger.error(f"Failed to access local data at {files_dir}", err, data={"local_store_dir": files_dir})
         de_notifier.failure()
-        raise Exception(
-            message.unexpected_error(f"Failed to access local data at {files_dir}", err)
-        ) from err
+
```
mikeAdamss: no need to lean a gap on any of these

dpypelines/pipeline/dataset_ingress_v1.py
```diff
@@ -62,30 +67,33 @@ def dataset_ingress_v1(files_dir: str, pipeline_config: dict):
                 )
                 raise ValueError(msg)
         except Exception as err:
+            logger.error(f"Error while looking for required file {required_file}", 
+            err, 
+            data={"required_file": required_file, 
+            "required_file_patterns": required_file_patterns, 
+            "local_store": local_store})
```
mikeAdamss: rather than passing in the store can we pass in a list of the files in the store? from memory there's a method to get them as a list, then just put it against a key of say "files_in_directory".

that way we've got the patterns we were looking for, the specific one that didn't match (from the error) and a list of the files considered.

dpypelines/pipeline/dataset_ingress_v1.py
```diff
@@ -34,25 +36,28 @@ def dataset_ingress_v1(files_dir: str, pipeline_config: dict):
...
-            message.unexpected_error("Failed to get required files patterns", err)
-        ) from err
+
+        raise err
 
     # Check for the existence of each required file
     for required_file in required_file_patterns:
         try:
             if not local_store.has_lone_file_matching(required_file):
+                logger.info(f"Required matching file {required_file} was not found.", data={"required_file": required_file})
```
mikeAdamss: I think you can get rid of most of this block. Id we dont' get a match just raise `FileNotFoundError(f"Could not find file matching pattern {required_file}")` and that will do. The error log after the catch will give us all the information we're after.

dpypelines/pipeline/dataset_ingress_v1.py
```diff
@@ -62,30 +67,33 @@ def dataset_ingress_v1(files_dir: str, pipeline_config: dict):
...
-            )
-        ) from err
+
+        raise err
 
     # Check for the existence of each supplementary distribution
     for supplementary_distribution in supplementary_distribution_patterns:
         try:
             if not local_store.has_lone_file_matching(supplementary_distribution):
+                logger.info(f"Supplementary distribution {supplementary_distribution} not found", data={"supplementary_distribution": supplementary_distribution})
```
mikeAdamss: see comments on required_files, same deal here

dpypelines/pipeline/dataset_ingress_v1.py
```diff
@@ -62,30 +67,33 @@ def dataset_ingress_v1(files_dir: str, pipeline_config: dict):
...
 
     # Extract the patterns for supplementary distributions from the pipeline configuration
     try:
         supplementary_distribution_patterns = (
             matching.get_supplementary_distribution_patterns(pipeline_config)
         )
+        logger.info("Successfully retrieved supplementary distribution patterns from pipeline config",
+         data={"supplementary_distribution_pattenrs": supplementary_distribution_patterns})
     except Exception as err:
+        logger.error("Failed to get supplementary distribution patterns", err, data={"local_store": local_store})
```
mikeAdamss: pass the config rather than the store please, supplementary distributions are in the config and that'll be the thing we need to know about if we cant get them.

mikeAdamss (commented):
a few tweaks but nothing major.

one thing - be careful about passing in `local_store` as a variable, its json logging so it'll just spit out the name of the class. It'd be better if you pass in `local_store.get_file_names()` - you'll end. up with a nice list of files within the directory in the logs.

dpypelines/pipeline/dataset_ingress_v1.py
```diff
@@ -95,57 +103,69 @@ def dataset_ingress_v1(files_dir: str, pipeline_config: dict):
...
-            ) from err
+            raise err
 
     # Get the positional arguments (the inputs) from the pipeline_config
     # dict and run the specified sanity checker for it
     args = []
     for match, sanity_checker in pipeline_config["transform_inputs"].items():
         try:
             input_file_path: Path = local_store.save_lone_file_matching(match)
+            logger.info(f"Successfully saved file that matches pattern to path {input_file_path}", data={"input_file_path": input_file_path})
```
mikeAdamss: good idea 👍 , but you dont need the variable in the message as you're capturing it with the data= kwarg.

I would probably also log out the match variable with the data kwarg too, so we can see at a glance which pattern resulted in which file being saved.

dpypelines/pipeline/dataset_ingress_v1.py
```diff
@@ -95,57 +103,69 @@ def dataset_ingress_v1(files_dir: str, pipeline_config: dict):
...
         args.append(input_file_path)
 
     # Get the transform function and keyword arguments from the transform_details
     transform_function = pipeline_config["transform"]
     kwargs = pipeline_config["transform_kwargs"]
+    logger.info("Retrieved transform function and keyword arguments from transform details.", data={"pipeline_config": pipeline_config})
 
     try:
         csv_path, metadata_path = transform_function(*args, **kwargs)
+        logger.info("Successfully retrieved csv path and metadata path with keyword args from transform function",
```
mikeAdamss: successfully _ran_ transform function, this is the DEs code running to completion. Then I'd maybe log out the csv. and metadata path with the data kwarg.

dpypelines/pipeline/dataset_ingress_v1.py
```diff
@@ -95,57 +103,69 @@ def dataset_ingress_v1(files_dir: str, pipeline_config: dict):
...
         csv_path, metadata_path = transform_function(*args, **kwargs)
+        logger.info("Successfully retrieved csv path and metadata path with keyword args from transform function", 
+        data={"transform_function": transform_function, 
+        "pipeline_config": pipeline_config})
+
     except Exception as err:
-        de_notifier.failure()
-        printable_transform_details = json.dumps(
-            pipeline_config, indent=2, default=lambda x: str(x)
+        logger.error("Error occured while getting transform function and keyword arguments from transform details",
```
mikeAdamss: error when running transform, include the name of the transform, args and kwargs with the data keyword.

dpypelines/pipeline/dataset_ingress_v1.py
```diff
@@ -34,118 +35,121 @@ def dataset_ingress_v1(files_dir: str, pipeline_config: dict):
...
-                msg = message.expected_local_file_missing(
-                    f"Supplementary distribution {supplementary_distribution} not found",
-                    supplementary_distribution,
-                    "dataset_ingress_v1",
-                    local_store,
-                )
-                raise ValueError(msg)
+                raise FileNotFoundError(f"Could not find file matching pattern {supplementary_distribution}")
         except Exception as err:
+            logger.error(f"Error while looking for supplementary distribution {supplementary_distribution}",
```
mikeAdamss, line 94: can you use the data= kwarg to inlcude the files_in_directory and config please. We need the same logic/reporting we applied to required_files.

dpypelines/pipeline/dataset_ingress_v1.py
```diff
@@ -34,118 +35,121 @@ def dataset_ingress_v1(files_dir: str, pipeline_config: dict):
...
-            ) from err
+            raise err
 
     # Get the positional arguments (the inputs) from the pipeline_config
     # dict and run the specified sanity checker for it
     args = []
     for match, sanity_checker in pipeline_config["transform_inputs"].items():
         try:
             input_file_path: Path = local_store.save_lone_file_matching(match)
+            logger.info("Successfully saved file that matches pattern", data={"input_file_path": input_file_path, "match": match})
```
mikeAdamss: can you include files_in_directory again please, so we saved X to Y from. a choice. of Z, that'd be everything you might want to know.

dpypelines/pipeline/dataset_ingress_v1.py
```diff
@@ -34,118 +35,121 @@ def dataset_ingress_v1(files_dir: str, pipeline_config: dict):
...
-                )
-            ) from err
+            raise err
 
         args.append(input_file_path)
 
     # Get the transform function and keyword arguments from the transform_details
     transform_function = pipeline_config["transform"]
     kwargs = pipeline_config["transform_kwargs"]
+    logger.info("Retrieved transform function and keyword arguments from transform details.", data={"pipeline_config": pipeline_config})
```
mikeAdamss: can you include data= fields of something like "args_got" and "kwargs_got", so we can see the config as well as the args and kwargs taken from it.

doesnt matter now I know, but at some point we'll likely be getting them via something a little less explicit that just calling a key, always good to have the before and after where we can.

mikeAdamss (commented):
minor comments.

mikeAdamss (approved):
LGTM 👍

PR #117, opened 2024-04-23
logging - s3 tar received

MuazzamChaud:
### What

Added a logger to each try and except to add additional information as to what code ran successfully and what failed, with appropriate variables to help debug when things go wrong.

### How to review

Look at each looger.info() and logger.error() and check:
- The message  is understandable and helps describes what's happening
- No spelling mistakes
- Check that data keyword is needed, where not used (and might not be needed, where used)
- Data keyword holds appropriate variables, and if any additional variables could be added
- Check to see where raw keyword might be useful
- Check to see if any additional loggers are needed anywhere

### Who can review

Anyone


nimshi89 (approved):
Seems fine to me

PR #119, opened 2024-04-23
Better acceptance test metadata feedback

nimshi89:
### What

Added the `diff` function form the `dictdiffer` library and the printing the results as a list. This way only the differences will be printed

The results are recorded before the assertion otherwise it did not print the list of differences. 
### How to review

Just run the behave test (as currently it is checking for a file that will have multiple differences) and this will be only till the ticket is approved.

### Who can review

anyone


features/steps/data.py
```diff
@@ -115,9 +115,10 @@ def step_impl(context, correct_metadata):
     correct_metadata_file = open(correct_metadata_path)
     correct_metadata_json = json.load(correct_metadata_file)
 
+    result = diff(context.json_output,correct_metadata_json)
     assert (
         context.json_output == correct_metadata_json
-    ), f"Metadata does not match expected metadata from {correct_metadata_json}."
+    ), f"Metadata does not match expected metadata, the following values are missing from the metadat output :\n {list(result)}."
```
SarahJohnsonONS: Add something to the error message to print out the full metadata output alongside the `diff` result, with info on interpreting the error (e.g. `add` means something is missing. `remove` means something needs deleting etc)

features/steps/data.py
```diff
@@ -106,7 +107,6 @@ def step_impl(context, metadata_output):
 
     context.json_output = json.load(metadata_file)
 
-
 @then("the metadata should match '{correct_metadata}'")
 def step_impl(context, correct_metadata):
```
SarahJohnsonONS, line 111: Add a "the metadata does not match {incorrect_metadata}" step

SarahJohnsonONS (commented):
Comments added

features/steps/data.py
nimshi89: Screenshot 2024-04-24 at 08 35 17
Changed the message and now with proper data it shows the path to the data that needs changing(or missing).

SarahJohnsonONS (approved):
LGTM

mikeAdamss:
closing, merged on a different branch for commit verification.

PR #121, opened 2024-04-24
remove confirmation to slack

mikeAdamss:
### What

Remove the "everything is ok" alarm where we punt details of the transform and metadata into slack.

Was only a sanity check until we had acceptance tests which we now do (and its spectacularly verbose now that we have metadata output).

I also removed an env var reference from the readme, not something thats used by this repo
 

### How to review

Sanity check

### Who can review

Anyone


Moasib-Arif (dismissed):
LGTM 👍

Moasib-Arif (approved):
LGTM

PR #126, opened 2024-04-30
email notifications for invalid data submission

Moasib-Arif:
### What

Added the email client to send emails if files not found or distribution file not found and for data submission.
Added feature flag the ability to turn the emails off
Created email templates

### How to review

Sanity check if the logic works and Check for any typos in utility.py, email_template_message.py and data_ingress_v1.py
check if Emails are sent from the s3_tar_recieved pipeline if you have AWS access.

AWS Email setup
In order to try this out you'll need to add the recipient email to ses and validate it (the usual confirmation url to an email address) - note, for the sender address I just created a free gmail email, you can do the same if you dont want to user a personal address, the setup steps once you have a recipient email are:

    sign in to the bleed env
    in the aws console go to simple email service
    click "identities"
    click email
    add the email you want to receive the emails
    go to the email you used and confirm the email address

set the env var TEMPORARY_SUBMITTER_EMAIL in glue.

    go to aws in the bleed env
    go to glue
    go to the pipeline
    look at the script tab - you can add temporary code there, you'll see the existing example,
    In the script change the email on Line 17 to the above verifed email.

Upload the Tar file to the s3 bucket and you will recieve an email regarding data submission if any files are missing you will recieve a seperate email.

### Who can review
Anyone with AWS Access


dpypelines/pipeline/dataset_ingress_v1.py
```diff
@@ -37,6 +44,27 @@ def dataset_ingress_v1(files_dir: str, pipeline_config: dict):
         logger.error("Error occured while attempting to create notifier instance", err)
         raise err
 
+    try:
+        submitter_email = get_submitter_email()
```
NickPapONS, line 48: Could you add logger statements (logger.info and logger.error) in these new try/except blocks? One for when something in the try block is done successfully, another for if it fails and there's an error. See other existing try/excepts in v1_data_ingress.py for reference, it would be a couple of lines only.

NickPapONS (changes requested):
Looks good, just a suggestion for the dataset_ingress_v1 additions

NickPapONS (approved):
Approved 👍

PR #129, opened 2024-05-03
Investigation and Implementation of SDMX 2.1 Transformation Script

osamede20:
### What

1. These SDMX 2.1  script was written to follow the same logic as that of SDMX 2.0.
This time, Python `ElementTree` package could not be applied as xml 2.1 file structure does not provide for easy accessibility of the element attributes. However, the SDMX 2.1 codes still deliver on performance, and the processing time is reasonable.

2. The `flatten_dict` from `utils.py` is called to get the unpacked header dictionary

3. The validation and tests from task DIS-714 are integrated into the code 

### How to review

The codes could be checked for performance optimisation 
They could also be checked for scalability.

### Who can review

Anybody


dpypelines/pipeline/shared/transforms/sdmx/generic/v21/prototype/v1.py
```diff
@@ -0,0 +1,302 @@
...
+                    observation_temp.update(
+                        {
+                            "generic:ObsValue": next(
+                                iter(observation["generic:ObsValue"].values())
+                            )
+                        }
+                    )
+                    obs_dicts.append(series_dict | observation_temp)
+            else:
+                print("something fucked up")
```
JimBryant7: I think the error message should be changed here.. not very descriptive :)

dpypelines/pipeline/shared/transforms/sdmx/generic/v21/prototype/v1.py
```diff
@@ -0,0 +1,302 @@
...
+    tables = data["message:GenericData"]["message:DataSet"]["generic:Series"]
+
+    series_dict = {}
+    obs_dicts = []
+
+    for table in tables:
+        for element in table.items():
+            if element[0] == "generic:SeriesKey":
+                df = pd.DataFrame(element[1]["generic:Value"])
+                series_dict = {a: b for a, b in zip(df["@id"], df["@value"])}
```
JimBryant7:
```
for sub_element in element[1]["generic:Value"]:
         series_dict[sub_element['@id']] = sub_element['@value']
```

replacing the two line with the above would be my suggestion

dpypelines/pipeline/shared/transforms/sdmx/generic/v21/prototype/v1.py
```diff
@@ -0,0 +1,302 @@
...
+            elif element[0] == "generic:Attributes":
+                df = pd.DataFrame(element[1]["generic:Value"])
+                series_dict.update({a: b for a, b in zip(df["@id"], df["@value"])})
```
JimBryant7:
```
for sub_element in element[1]["generic:Value"]:
               series_dict[sub_element['@id']] = sub_element['@value']
```

replacing the two line with the above would be my suggestion

dpypelines/pipeline/shared/transforms/sdmx/generic/v21/prototype/v1.py
```diff
@@ -0,0 +1,302 @@
...
+            elif element[0] == "generic:Obs":
+                for observation in element[1]:
+                    observation_temp = {}
+                    df = pd.DataFrame(
+                        observation["generic:Attributes"]["generic:Value"]
+                    )
+                    observation_temp.update(
+                        {a: b for a, b in zip(df["@id"], df["@value"])}
+                    )
```
JimBryant7:
```
for sub_element in observation["generic:Attributes"]["generic:Value"]:
             observation_temp[sub_element['@id']] = sub_element['@value']
```

replacing the 6 lines with the above would be my suggestion

JimBryant7 (commented):
So my suggestions are based around removing dataframes whilst iterating. In other words pandas dataframes are very inefficient when being used whilst iterating. So I have removed them and just added the required data straight into the series/observation dictionaries (using a for loop). I've run the transform with these changes and it runs significantly quicker.

dpypelines/pipeline/shared/transforms/sdmx/generic/v21/prototype/v1.py
```diff
@@ -0,0 +1,302 @@
...
+    )  # transform validation
+
+    header_dict = flatten_dict(header)
+    check_header_unpacked(header_dict)  # transform validation
+    header_df = pd.DataFrame([header_dict])
+
+    repl_rows = header_df.loc[0].copy()
+    header_frame = pd.concat(
+        [header_df, pd.DataFrame([repl_rows] * (len(obs_frame) - 1))], ignore_index=True
+    )
```
JimBryant7: I've found that 
```
pd.DataFrame([repl_rows] * (len(obs_frame) - 1))]
```
is also quite inefficient. I have a suggestion to replace it with, it may seem pedantic but I'm just concerned that for much larger datasets it will slow it down unnecessarily. My suggestion is to replace the above 4 lines with
```
header_dict_2 = header_df.to_dict(orient='list')
for key in header_dict_2:
        header_dict_2[key] = header_dict_2[key] * (len(obs_frame) - 1)
    
header_frame_2 = pd.DataFrame(header_dict_2)
header_frame = pd.concat(
        [header_df, header_frame_2], ignore_index=True
    )
```
essentially you are using a dict and duplicating items in a list to the required length, rather than duplicated a pd.Series (again its pandas that is inefficient) 
(also my variable names are quite lazy ahah so change if you wish)

osamede20:
All suggested changes have been made and there is a remarkable improvement in the performance of the transform code.
Thanks.

JimBryant7 (approved):

PR #130, opened 2024-05-03
validation and testing for sdmx 2.1 transform

osamede20:
### What

1. Created transform validation functions and unit tests for SDMX 2.1. 
2. Implemented validation functions into the SDMX 2.1 transform to ensure the transform script was correctly written and the output tidy csv is correct.
3. Implemented data quality metrics by monitoring for anomalies, outliers, missing values, or unexpected duplicates etc. to ensure the integrity and accuracy of the data.

### How to review

1. Check the validations to see that they make sense and working correctly. 
2. Check unit tests have passed using make test. 
3. Possibly, try the SDMX 2.1 transform out on a xml 2.1 file.

### Who can review

Any team member can review

### DIFF
--- a/dpypelines/pipeline/shared/transforms/validate_transform_v21.py
+++ b/dpypelines/pipeline/shared/transforms/validate_transform_v21.py
@@ -1,0 +1,94 @@
+"""
+Transform validation functions to be used by dpypelines.pipeline.shared.transforms.sdmx.generic.v21.prototype.v1
+ie -  generic sdmx data version 2_1
+file may need renaming when further transforms for different sdmx's are created
+"""
+
+from pathlib import Path
+
+import pandas as pd
+
+
+def check_read_in_sdmx(xml_file: Path):
+    # check sdmx can be read in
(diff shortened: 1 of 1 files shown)


JimBryant7 (commented):
I feel there should be a test that makes sure that each dict of obs_dicts has the same keys, just to make sure nothing is missing from each series block.

Also I think it is worth writing unit tests for these functions, I can see a lot of these are the same as the validations for the other version of sdmx so copying a lot of them over will save time :)

osamede20:
Yes, there should be a test that checks the `obs_dicts` (which is actually a list in sdmx 2.1 transform) has all the keys in the series block. Yes, unit tests for the functions would be good. It would take some time to figure out the logic of getting the keys from the `series_dicts` on the raw sdmx 2.1 file.

To get this ticket over the line for this Sprint, perhaps, we could review this and sort out the suggested tests in the next Sprint. 
Thanks.

JimBryant7 (approved):

PR #131, opened 2024-05-10
Functions added to get transform details from config

SarahJohnsonONS:
### What

Three functions added (in `dpypelines/pipeline/shared/pipelineconfig/transform.py`) to get transform inputs, transform function and transform kwargs from pipeline config dict. `dataset_ingress_v1()` updated to use new functions instead of hardcoded values. Associated tests added in `tests/pipelines/pipeline/shared/pipelineconfig/test_transform.py`. Tided up some logging statements in `dataset_ingress_v1()`.

### How to review

Sanity check, make sure `dataset_ingress_v1()` runs up to L321.

### Who can review

Anyone.

### DIFF
--- a/dpypelines/pipeline/dataset_ingress_v1.py
+++ b/dpypelines/pipeline/dataset_ingress_v1.py
@@ -6,7 +6,6 @@ from dpytools.http.upload import UploadServiceClient
 from dpytools.logging.logger import DpLogger
 from dpytools.stores.directory.local import LocalDirectoryStore
 
-from dpypelines.pipeline.shared import message
 from dpypelines.pipeline.shared.email_template_message import (
     file_not_found_email,
     submission_processed_email,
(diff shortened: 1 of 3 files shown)


dpypelines/pipeline/dataset_ingress_v1.py
```diff
@@ -211,15 +210,16 @@ def dataset_ingress_v1(files_dir: str, pipeline_config: dict):
...
-                    submitter_email,
-                    " about missing supplementary distribution: ",
-                    supp_dist_pattern,
+                    "Email sent to submitter about missing supplementary distribution file",
+                    data={
+                        "submitter_email": submitter_email,
+                        "supplementary_distribution_pattern": supp_dist_pattern,
+                    },
                 )
                 de_notifier.failure()
```
mikeAdamss, line 223: wasn't you but but there's a logic error here. A missing file is an error not information, we need the pipeline to error and stop processing.

Can you (a) create an error (b) log it, then (c) raise it after we notify of the failure please.

you can just instantiate an error manually, it wont raise in of itself, something like:

```
err = FileNotFoundError(something useful>)
```

mikeAdamss (commented):
code is good but spotted something else while I was i there, quick fix.

mikeAdamss (approved):
LGTM

PR #135, opened 2024-05-17
709-remove outdated approach

MuazzamChaud:
### What

Removing outdated code such as anything referencing schemas (including tests) and messages.

### How to review

Check that the removed code/files are in fact outdated.

### Who can review

Anyone


Moasib-Arif (approved):
LGTM 👍

PR #137, opened 2024-05-20
Tests passing ready for review

NickPapONS:
### What

Added a new acceptance test that runs the pipeline similarly to the existing "Pipeline runs without errors" test, but using the sdmx 2.1 transform functionality.

### How to review

Try the behaviour tests to see if they pass, check the test code to see if it makes sense.

### Who can review

Anyone


features/steps/dataset_ingress.py
```diff
@@ -38,6 +39,37 @@
         "secondary_function": dataset_ingress_v1,
     },
 }
+
+CONFIGURATION_SDMX_2_1 = {
```
SarahJohnsonONS: You could just add a new key to the existing configuration dictionary, e.g. `valid_no_supp_dist_v2_1` and change the other key to `valid_no_supp_dist_v2_0`. You could get rid of the separate step definitions for getting dataset IDs, in that case (on lines 102 and 107)

NickPapONS: Good idea, not sure why I split it all up so much. I'll give it a try now

PR #138, opened 2024-05-20
Bump dp python tools

mikeAdamss:
### What

bump in use dp-python-tools to version 0.1.1. I've also removed the florence token logic from the pipelines, this is all now handled on the upload client side.

changed the acceptance test order slightly to better reflect the order of actions (confirm the things are right before sending them somewhere).

also moved the success notification to the end of the script as we were reporting success a little early.

### How to review

Sanity check.

### Who can review

Anyone.


Moasib-Arif (approved):
LGTM 👍

PR #140, opened 2024-05-21
readme for runing dataset_ingress_v1, plus SKIP_DATA_UPLOAD capability

mikeAdamss:
### What

- Added instructions for running `dataset_ingress_v1.py`.
- Added a new env var of `SKIP_DATA_UPLOAD` as we probaly dont want to upload files to s3 every time we run a transform we're developing.
- Fixed a mistake of mine where I said we dont need to raise errors to log them, we do, the logger needs the stack trace.

### How to review

Sanity check - **run the script** as per instructions and make sure it works for you.

### Who can review

Anyone.

### DIFF
--- a/dpypelines/pipeline/README.md
+++ b/dpypelines/pipeline/README.md
@@ -4,33 +4,58 @@ Instructions on running existing pipelines can be found below. As more pipeline 
 
 ## dataset_ingress_v1
 
-### Default configuration
+### Running locally
 
-The `dataset_ingress_v1` pipeline function can be run directly with a local directory of files using the default configuration details.
+The `dataset_ingress_v1` pipeline function can be run directly via a script like the following:
 
+`myscript.py`
 ```python
 from dpypelines.pipeline.configuration import get_pipeline_config
 from dpypelines.pipeline.dataset_ingress_v1 import dataset_ingress_v1
 
-# `files_dir` should be a string of the directory containing a data.xml file
-files_dir = "path/to/data/file"
+# `files_dir` should be a string of the directory containing your data source(s)
+files_dir = "path/to/directory/of/inputs"
 
-# Get the default configuration details
-default_pipeline_config, _ = get_pipeline_config("default")
+# Specify the dataset id for picking the configuration you want to use for
+# your source.
+default_pipeline_config  = get_pipeline_config("<dataset_id>")
 
+# Run the function
 dataset_ingress_v1(files_dir, default_pipeline_config)
 ```
 
-Running `dataset_ingress_v1` will generate three files:
-- a copy of the source data.xml file.
+Which you would run via:
+
+```
+poetry run python3 ./myscript.py
+```
+
+Note: the key you choose should match at least one key in the `CONFIGURATION` dict, see [this file](https://github.com/ONSdigital/dp-data-pipelines/blob/sandbox/dpypelines/pipeline/configuration.py).
+
+**Before** running this, you'll need to set some environment variables based on one of two scenarios as follows:
+
+### Scenario 1: You dont care about uploading the results of the pipeline (you just want to run the transform and see the outputs)
+
+This is probably the setup you'll want to use the majority of the time when developing locally.
+
+Set the following env vars:
+
+```
+export DISABLE_NOTIFICATIONS=true
+export SKIP_DATA_UPLOAD=true
+```
+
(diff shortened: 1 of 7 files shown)


dpypelines/pipeline/dataset_ingress_v1.py
```diff
@@ -345,94 +349,127 @@ def dataset_ingress_v1(files_dir: str, pipeline_config: dict):
...
+        elif skip_data_upload.strip() in ["False", "false"]:
+            skip_data_upload = False
+        else:
+            # Catch a trivial raise as we need the stack trace of the
+            # error for the logger, so it needs to be a raised error.
+            try:
+                raise ValueError(
+                    "If present SKIP_DATA_UPLOAD must be one of: True, true, False, false."
+                )
+            except ValueError as err:
```
SarahJohnsonONS: Should this be `ValueError` or `Exception`?

dpypelines/pipeline/README.md
```diff
@@ -4,33 +4,58 @@ Instructions on running existing pipelines can be found below. As more pipeline
...
+**Before** running this, you'll need to set some environment variables based on one of two scenarios as follows:
+
+### Scenario 1: You dont care about uploading the results of the pipeline (you just want to run the transform and see the outputs)
+
+This is probably the setup you'll want to use the majority of the time when developing locally.
+
+Set the following env vars:
+
+```
+export DISABLE_NOTIFICATIONS=true
```
SarahJohnsonONS, line 44: Need to set UPLOAD_SERVICE_URL as well (can be a nonsense string if not actually uploading)

dpypelines/pipeline/dataset_ingress_v1.py
```diff
@@ -187,25 +187,29 @@ def dataset_ingress_v1(files_dir: str, pipeline_config: dict):
...
-                        "submitter_email": submitter_email,
-                        "supplementary_distribution_pattern": supp_dist_pattern,
-                    },
-                )
-                de_notifier.failure()
-                raise err
+                # Catch a trivial raise as we need the stack trace of the
+                # error for the logger, so it needs to be a raised error.
+                try:
+                    raise FileNotFoundError(
```
SarahJohnsonONS, line 197: Do we want to send an email at this point?

dpypelines/pipeline/dataset_ingress_v1.py
```diff
@@ -345,94 +349,127 @@ def dataset_ingress_v1(files_dir: str, pipeline_config: dict):
...
-    except Exception as err:
-        logger.error(
-            "Error creating UploadClient", err, data={"upload_url": upload_url}
-        )
-        de_notifier.failure()
-        raise err
+    # Allow DE's to skip the upload to s3 part of the pipeline while
+    # developing code locally.
+    skip_data_upload = os.environ.get("SKIP_DATA_UPLOAD", False)
+    if skip_data_upload is not False:
```
SarahJohnsonONS, line 359: Use `dpypelines.pipeline.shared.utils.str_to_bool` here?

dpypelines/pipeline/dataset_ingress_v1.py
mikeAdamss: we sent if just below. By no means a hill I'll die on.

this is the bit I wrong about before - we need to have a raised error (so the logger can create info from the stack trace).

so we:
- raise it
- catch it
- log it
- send email
- raise it properly.

you could email just before here like you say, then raise and catch it to do the logger part, but seemed slightly neater having just the raise. By no means a hill I'd die on.

did that make sense or am I misunderstanding you?

dpypelines/pipeline/dataset_ingress_v1.py
mikeAdamss: ty, I literally just added a card to add that function, forgot we made one 😆

dpypelines/pipeline/dataset_ingress_v1.py
mikeAdamss: ValueError should be fine, its nice and explicit. Lets keep it handling just the scenario we're creating.

dpypelines/pipeline/dataset_ingress_v1.py
SarahJohnsonONS: Makes sense, I've just never seen the `try: raise, except: raise again` structure before.

SarahJohnsonONS (approved):
LGTM

PR #141, opened 2024-05-23
709 remove outdated approach verified commints branch

MuazzamChaud:
### What

Remove files and code referring to outdated approach

### How to review

Check removed code is actually outdated

### Who can review

Anyone


Moasib-Arif (approved):
LGTM 👍

PR #145, opened 2024-06-05
Create generic_file_ingress_v1.py

osamede20:
### What

https://jira.ons.gov.uk/browse/DIS-844

1. Created the `generic_file_ingress_v1.py` for files that need moving only and do not need transform.
2. Included a configuration key on `configuration.py` for files that need moving only

### How to review

1. Check the codes for functionality and logic, to see whether the scripts work correctly.
2. Check whether the `generic_file_ingress_v1.py` correctly moves or uploads sdmx files.

### Who can review

Any team member


dpypelines/pipeline/configuration.py
```diff
@@ -23,13 +25,22 @@
...
         "config_version": 1,
-        "transform": sdmx_compact_2_0_prototype_1,
+        "transform": sdmx_generic_2_1_prototype_1,
         "transform_inputs": {"^data.xml$": sdmx_sanity_check_v1},
         "transform_kwargs": {},
         "required_files": [{"matches": "^data.xml$", "count": "1"}],
         "supplementary_distributions": [{"matches": "^data.xml$", "count": "1"}],
         "secondary_function": dataset_ingress_v1,
     },
+    "*._move$": {
```
mikeAdamss: you need to understand how the regex logic of this config works.

here's a few super basic regex meanings:

`^` -  start of string
`$` - end of string
`.*` -  any number of any characters.

therefore this is what each of which is regex statements is:

`^cpih$` - string that starts has the text "cpih" then ends, i.e has no characters to either side of cpih.
`^.*$"` - string that starts, contains any number of any characters then ends.
`^*._move$` -  string that starts, contains any number of any characters, then "_move" then ends.
 
 the **ordering** is what's wrong here. When we look for a match we look for t**he first pattern than matches the id**. So right now any source/dataset id you use will match `^.*$` so will never get compared to `*._move$`.
 
 the trick is going to be always order the keys of this dictionary with the most specific at the top and the least specfici/broadest at the bottom.
 
 have a think but I'd imagine the keys would be something like this:
 
 `^.*_sdmx_v2_1$` - any source id ending in _sdmx_2_1 gets the config for handling sdmx versions 2.1 submissions
 ......if the source id does not match the above....
 `^.*_sdmx_v2_0$` - any source id ending in sdmx_v2_0 gets the config for handling sdmx versions 2.0 submissions 
 ......if the source id does not match the above....
 `^.*_move$` - any source id ending in _move gets the new config for just moving the file without transform.
 
 IF (and only id) you wanted a default fallback you'd have a `^.*$` key as the very last thing, so the generic default. Though tbh I doubt we want that.

dpypelines/pipeline/generic_file_ingress_v1.py
```diff
@@ -0,0 +1,315 @@
...
+    # Create email client from env var
+    try:
+        email_client = get_email_client()
+        logger.info("Created email client", data={"email_client": email_client})
+    except Exception as err:
+        logger.error("Error occurred when creating email client", err)
+        de_notifier.failure()
+        raise err
+
+    # just throw out an email to see if it works
```
mikeAdamss: can you delete this try catch block. Sending an email for no purpose was code that slipped through review (we dont want to email someone for no reason). It's already deleted in the main pipeline,

dpypelines/pipeline/generic_file_ingress_v1.py
```diff
@@ -0,0 +1,315 @@
...
+        logger.error(
+            "Error occurred when getting supplementary distribution patterns",
+            err,
+            data={"pipeline_config": pipeline_config},
+        )
+        de_notifier.failure()
+        raise err
+
+    # Check for the existence of each supplementary distribution
+    for supp_dist_pattern in supp_dist_patterns:
```
mikeAdamss: remove please, supplementary_distributions will be blank.

dpypelines/pipeline/generic_file_ingress_v1.py
```diff
@@ -0,0 +1,315 @@
...
+            )
+        except Exception as err:
+            logger.error(
+                "Error creating UploadClient", err, data={"upload_url": upload_url}
+            )
+            de_notifier.failure()
+            raise err
+
+        # Check for supplementary distributions to upload
+        if supp_dist_patterns:
```
mikeAdamss: dont check for existence, we're already confirmed we have 1 or more required_files - just upload them

dpypelines/pipeline/configuration.py
```diff
@@ -23,13 +25,22 @@
...
+        "config_version": 1,
+        "transform": None,
+        "transform_inputs": {},
+        "transform_kwargs": {},
+        "required_files": [{"matches": "^data.xml$", "count": "1"}],
```
mikeAdamss: the only file we actually required is exactly 1 file thats **not** called manfiest.json.

NOT matches with regex are a thing. You'll need to do some googling around and use the regex playground I linked but you can do it.

dpypelines/pipeline/configuration.py
mikeAdamss: there is a nice little site here that lets you play with and try out regex statments: https://regexr.com/

dpypelines/pipeline/configuration.py
```diff
@@ -23,13 +25,22 @@
...
+        "supplementary_distributions": [{"matches": "^data.xml$", "count": "1"}],
```
mikeAdamss: supplementary_distributions is for files to upload alongside outpuuts from transforms. There's no transform, this can be blank.

dpypelines/pipeline/generic_file_ingress_v1.py
```diff
@@ -0,0 +1,315 @@
...
+                    "required_file": required_file,
+                    "required_file_patterns": required_file_patterns,
+                    "files_in_directory": files_in_directory,
+                    "pipeline_config": pipeline_config,
+                },
+            )
+            de_notifier.failure()
+            raise err
+
+    # Extract the patterns for supplementary distributions from the pipeline configuration
```
mikeAdamss: remove please, supplementary_distributions will be blank.

dpypelines/pipeline/shared/transforms/sdmx/v20.py
```diff
@@ -0,0 +1,71 @@
...
+
+import pandas as pd
+
+from dpypelines.pipeline.shared.transforms.sdmx.compact.v20.prototype.v1 import (
+    generate_versions_metadata,
+    xmlToCsvSDMX2_0,
+)
+
+
+def sdmx_compact_2_0_prototype_1(input_file: Path):
```
mikeAdamss, line 9: assuming im following what you're doing here, you're also going to need an. sdmx_compact_2_1_prototype_1 function.

mikeAdamss (commented):
left some comments that I hope make sense.

you've also got a failing test or check of some sort at the moment (if you look on the main pr page, you'll see a red X nearer the bottom). Try and make sure those are all green ticks if you can please.

dpypelines/pipeline/configuration.py
osamede20: Thanks.
I am not sure I completely understand what to do here. However, I have kept the required file to match `data.xml` as this is the input file with which I tested the `generic_file_ingress_v1.py `

dpypelines/pipeline/configuration.py
osamede20: Many thanks

dpypelines/pipeline/generic_file_ingress_v1.py
osamede20: done.

dpypelines/pipeline/generic_file_ingress_v1.py
osamede20: done.

dpypelines/pipeline/generic_file_ingress_v1.py
osamede20: done

dpypelines/pipeline/configuration.py
osamede20: done

dpypelines/pipeline/generic_file_ingress_v1.py
osamede20: done

dpypelines/pipeline/shared/transforms/sdmx/v20.py
osamede20: yes, i had created another function `sdmx_generic_2_1_prototype_1` that takes care of the transform for sdmx v2.1

dpypelines/pipeline/generic_file_ingress_v1.py
```diff
@@ -0,0 +1,184 @@
...
+from dpypelines.pipeline.shared.pipelineconfig.matching import (
+    get_required_files_patterns,
+)
+from dpypelines.pipeline.shared.utils import (
+    get_email_client,
+    get_submitter_email,
+    str_to_bool,
+)
+
+logger = DpLogger("generic-ingress-pipeline-v1")
```
mikeAdamss: can you change this to just "data-ingress-pipelines" here **and** in dataset_ingress_v1 please.

think of it as the heading that logs will appear under. We probably want all the logs for pipelines under the same heading.

dpypelines/pipeline/generic_file_ingress_v1.py
```diff
@@ -0,0 +1,184 @@
...
+            "Required file patterns retrieved from pipeline config",
+            data={"required_file_patterns": required_file_patterns},
+        )
+    except Exception as err:
+        files_in_directory = local_store.get_file_names()
+        logger.error(
+            "Error occurred when getting required file patterns",
+            err,
+            data={
+                "files_in_directory": files_in_directory,
```
mikeAdamss: you don't need this information. There's an error getting the required file patterns from the config. The. files that happen to be in the directory aren't relevant to that and might confuse people.

could you also remove that data field from dataset_ingress_v1 please.

dpypelines/pipeline/generic_file_ingress_v1.py
```diff
@@ -0,0 +1,184 @@
...
+                    submitter_email, email_content.subject, email_content.message
+                )
+                logger.info(
+                    "Email sent to submitter about missing required file",
+                    data={
+                        "submitter_email": submitter_email,
+                        "required_file": required_file,
+                    },
+                )
+                de_notifier.failure()
```
mikeAdamss: we should raise an error here. We've found out that a required file is missing and notified of a failure but we're not rising an error so the pipeline is just gonna keep going.

**After** the zde_notifier.failure()` can you add something like

```
raise FileNotFoundError(f"Could not find file matching pattern {required_file}")
```
Again its wrong in both pipelines so you'll need to correct in both sorry.

dpypelines/pipeline/generic_file_ingress_v1.py
```diff
@@ -0,0 +1,184 @@
...
+    BasePipelineNotifier,
+    notifier_from_env_var_webhook,
+)
+from dpypelines.pipeline.shared.pipelineconfig.matching import (
+    get_required_files_patterns,
+)
+from dpypelines.pipeline.shared.utils import (
+    get_email_client,
+    get_submitter_email,
+    str_to_bool,
```
mikeAdamss: str_to_bool was recently moved to a different place. You'll need to merge in main. You can now import in from dpytools like this: `from dpytools.utilities.utilities import str_to_bool`

dpypelines/pipeline/generic_file_ingress_v1.py
```diff
@@ -0,0 +1,184 @@
...
+        "skip_data_upload set from SKIP_DATA_UPLOAD env var",
+        data={"value": skip_data_upload},
+    )
+
+    if skip_data_upload is not True:
+
+        # Upload output files to Upload Service
+        try:
+            # Create UploadClient from upload_url
+            UploadServiceClient(upload_url)
```
mikeAdamss: `upload_client = UploadServiceClient(upload_url)`

dpypelines/pipeline/generic_file_ingress_v1.py
```diff
@@ -0,0 +1,184 @@
...
+            logger.info(
+                "UploadClient created from upload_url", data={"upload_url": upload_url}
+            )
+        except Exception as err:
+            logger.error(
+                "Error creating UploadClient", err, data={"upload_url": upload_url}
+            )
+            de_notifier.failure()
+            raise err
+
```
mikeAdamss, line 243: the above block just creates `upload_service_client` (with the tweak I've put in a comment).

now you have an instantiated upload_client you need to:

- loop through all the require files
- ignore/continue if the file iis called "manifest .json"
- otherwise upload it with the upload client. 
       - if its an xml file, then you'd need to use `upload_client.upload_new_sdmx(<path to file>)`
       - if its a csv file, then you'd need to use `upload_client.upload_new_csv(<path to file>)`
       - any other type then (for now) you;'ll need to raise an error as these are the currently supported file types. 
       
theres some documentation on the uplaod client here: https://github.com/ONSdigital/dp-python-tools/tree/develop/dpytools/http. You onyl want the methods with "new" in them.

dpypelines/pipeline/configuration.py
mikeAdamss: You;'ll need to check this in that regex platground, but I think what you're looking for is:

```
^(?!manifest\.json$)
```

Which I think matches any file name thats NOT exactly equal to manifest.json.

That's the requirment for this pipeline. Exactly 1 file of any name alongside the manfiest.json file.

dpypelines/pipeline/shared/transforms/sanity_check.py
```diff
@@ -1,15 +1,40 @@
 from pathlib import Path
 
-import pandas as pd
+
+def sdmx_sanity_check_v1_file_extension(sdmx_file: Path):
```
mikeAdamss: if you're using this I think you'll need a config that specifies that you're using this. You specify the sanity checker you;'re using for a given input in the config in the "transform_inputs" field.

dpypelines/pipeline/shared/transforms/sanity_check.py
```diff
@@ -1,15 +1,40 @@
...
     try:
-        pd.read_xml(sdmx_file)
+        with open(sdmx_file, "r") as f:
+            f.read()
+        assert sdmx_file.name.endswith(".xml"), "Invalid sdmx_file"
     except Exception as err:
         raise Exception(f"Failed to read in xml - {sdmx_file}") from err
+
+
+# def sdmx_sanity_check_v1(sdmx_file: Path):
```
mikeAdamss: if we're not using it, can we delete it please.

mikeAdamss (commented):
Added some comments.

This is coming together nicely but you need to get the upload client stuff in here.

dpypelines/pipeline/configuration.py
osamede20: Thanks; done!

dpypelines/pipeline/generic_file_ingress_v1.py
osamede20: Thanks; done!

dpypelines/pipeline/generic_file_ingress_v1.py
osamede20: Thanks.
`"files_in_directory": files_in_directory` removed from the data field.
The data field is kept in the code with the name-value `"pipeline_config": pipeline_config` incase we need to see error details in the config file

dpypelines/pipeline/generic_file_ingress_v1.py
osamede20: Done

dpypelines/pipeline/generic_file_ingress_v1.py
osamede20: Done

dpypelines/pipeline/generic_file_ingress_v1.py
osamede20: Thanks; done

dpypelines/pipeline/generic_file_ingress_v1.py
osamede20: Thanks; done

dpypelines/pipeline/shared/transforms/sanity_check.py
osamede20: Thanks.
`sdmx_sanity_check_v1(sdmx_file: Path)` is specifed in the "transform_input" field in the configuration.py.

A sanity check that the recieved sdmx file ends with `".xml"` is also included in  `sdmx_sanity_check_v1(sdmx_file: Path)`  as it is in `sdmx_sanity_check_v1_file_extension(sdmx_file: Path)`. So, I guess having the function `sdmx_sanity_check_v1_file_extension(sdmx_file: Path)` on `sanity_check.py` is probably redundant or a repetition.

dpypelines/pipeline/shared/transforms/sanity_check.py
osamede20: Done

dpypelines/pipeline/configuration.py
```diff
@@ -1,16 +1,17 @@
...
+from dpypelines.pipeline.shared.transforms.sdmx.v21 import sdmx_generic_2_1_prototype_1
 
 # Set a regex pattern matching the `source_id` as `CONFIGURATION` dictionary key
 # All fields are required in order for a pipeline transform to run successfully
 CONFIGURATION = {
-    # This is an example of how to set configuration details for a specific source_id (`cpih` in this case)
-    "^cpih$": {
+    # This is an example of how to set configuration details for a generic source_id - ending with v2_0, v2_1 for sdmx 2.0 and 2.1 respectively, and
+    # generic source_id - ending with move for an sdmx file to be moved
+    "^.*_v2_0$": {
```
mikeAdamss: you probably need to have an indicator that its sdmx and compact in here.

- what if its not compact, would it not need a different function?
- what if its v2_0 of something else? maybe we start using a different data format and end up with a source_id of new_cool_format_v2_0 - that'd end up getting processed as sdmx right?

dpypelines/pipeline/configuration.py
```diff
@@ -19,17 +20,24 @@
         "supplementary_distributions": [{"matches": "^data.xml$", "count": "1"}],
         "secondary_function": dataset_ingress_v1,
     },
-    # Default configuration (regex pattern matches any string of characters of length >= 0)
-    # This *always* needs to be the final entry in the dictionary to prevent inadvertent matches
-    "^.*$": {
+    "^.*_v2_1$": {
```
mikeAdamss: same as other comment, version 2.1 of what format?

dpypelines/pipeline/generic_file_ingress_v1.py
```diff
@@ -0,0 +1,199 @@
...
+        # Upload output files to Upload Service
+        try:
+            # Create UploadClient from upload_url
+            upload_client = UploadServiceClient(upload_url)
+            for required_file in required_file_patterns:
+                if "manifest.json" in required_file:
+                    continue
+                elif ".xml" in required_file:
+                    upload_client.upload_new_sdmx(files_dir)
```
mikeAdamss: files_dir is the entire directory, you need to be telling it to upload the **specific** file that matches the required_file that's not manifest.json.

you'll need to save the file matching the pattern to get the path, then pass that path into the upload method, see here: https://github.com/ONSdigital/dp-data-pipelines/blob/6f13dd3c7e992cf4913d7d8a31fc41cc124cdce0/dpypelines/pipeline/dataset_ingress_v1.py#L424

dpypelines/pipeline/generic_file_ingress_v1.py
```diff
@@ -0,0 +1,199 @@
...
+                    logger.info(
+                        "UploadClient created from upload_url",
+                        data={"upload_url": upload_url},
+                    )
+                elif ".csv" in required_file:
+                    upload_client.upload_new_csv(files_dir)
```
mikeAdamss: see other comment, upload the specific required file, not the directory.

dpypelines/pipeline/generic_file_ingress_v1.py
```diff
@@ -0,0 +1,199 @@
...
+                    logger.info(
+                        "UploadClient created from upload_url",
+                        data={"upload_url": upload_url},
+                    )
+                else:
+                    raise TypeError(
+                        f"File type of {required_file} currently not supported "
+                    )
+        except Exception as err:
+            logger.error(
```
mikeAdamss, line 192: this isnt just the errror for creating the client, It's any error generated by all the permutations of the upload logic. You need to break this out.

- 1. Put creating the upload client in its own try catch block with this error catch and log.
- 2. Put the loop and upload logic in a different try catch block, when you log errors from _that_ you can then include the name of the required_file that failed to upload.

dpypelines/pipeline/shared/transforms/sanity_check.py
mikeAdamss: it looks redundant to me. I'd delete it if its not being called.

dpypelines/pipeline/shared/transforms/sanity_check.py
```diff
@@ -1,15 +1,23 @@
...
     Sanity check that the recieved sdmx file is actually an sdmx file.
     """
+
     assert sdmx_file.name.endswith(".xml"), "Invalid sdmx_file"
 
     try:
-        pd.read_xml(sdmx_file)
+        with open(sdmx_file, "r") as f:
+            f.read()
+        assert sdmx_file.name.endswith(".xml"), "Invalid sdmx_file"
```
mikeAdamss: we're already doing this on line 16

dpypelines/pipeline/shared/transforms/sdmx/v20.py
```diff
@@ -0,0 +1,66 @@
...
+    csv_out = Path("data.csv")
+    metadata_out = Path("metadata.json")
+
+    xmlToCsvSDMX2_0(input_file, csv_out)
+    generate_versions_metadata(csv_out, metadata_out)
+
+    return csv_out, metadata_out
+
+
+def stub_smdx_default_v1():
```
mikeAdamss: we've defined theses stubbed functions on both `./v20.py` and `./v21.py`. I would move the stub functions to say `./stub.py` and import them in v20 and v21, otherwise we're repeating ourselves.

dpypelines/pipeline/shared/transforms/sdmx/v21.py
```diff
@@ -0,0 +1,62 @@
...
+    csv_out = Path("data.csv")
+    metadata_out = Path("metadata.json")
+
+    xmlToCsvSDMX2_1(input_file, csv_out)
+    generate_versions_metadata(csv_out, metadata_out)
+
+    return csv_out, metadata_out
+
+
+def stub_smdx_default_v1():
```
mikeAdamss: see comment in v20.py

mikeAdamss (commented):
few more comments, nothing major

dpypelines/pipeline/generic_file_ingress_v1.py
osamede20: done.

dpypelines/pipeline/generic_file_ingress_v1.py
osamede20: done.

dpypelines/pipeline/generic_file_ingress_v1.py
osamede20: done.

dpypelines/pipeline/shared/transforms/sanity_check.py
osamede20: done

dpypelines/pipeline/shared/transforms/sdmx/v20.py
osamede20: thanks; done!

dpypelines/pipeline/shared/transforms/sdmx/v21.py
osamede20: thanks; done!

nimshi89:
I had a look and it seems fine for me, but some one more experienced should approve this PR.

RedWalters:
Can't find any further issues, approved

RedWalters (dismissed):

dpypelines/pipeline/generic_file_ingress_v1.py
```diff
@@ -0,0 +1,257 @@
...
+        manifest_dict = local_store.get_lone_matching_json_as_dict("manifest.json")
+        logger.info(
+            "Got manifest.json dict output",
+            data={"manifest_dict": manifest_dict},
+        )
+    except Exception as err:
+        logger.error(
+            "Error occurred when getting manifest_dict",
+            err,
+            data={"manifest_dict": manifest_dict},
```
SarahJohnsonONS: Remove the `data` parameter here as there won't be a `manifest_dict` to display

dpypelines/pipeline/generic_file_ingress_v1.py
```diff
@@ -0,0 +1,257 @@
...
+                    logger.error(
+                        "Error uploading file.",
+                        err,
+                        data={"file_path": required_file_path},
+                    )
+                    de_notifier.failure()
+                    raise err
+            else:
+                # TODO add logging.error
+                try:
```
SarahJohnsonONS: Replace the `else: try...except` block (L243-L256) with:
```
else:
    raise NotImplementedError(f"Uploading file type {required_file_path.suffix} not currently supported.")
```

SarahJohnsonONS (commented):
A couple of tiny changes then it'll be ready for approval

SarahJohnsonONS (approved):
LGTM :)

PR #146, opened 2024-06-06
logging comes from nopemailclient

MuazzamChaud:
### What

Removed logging about sent emails in pipeline, and now NopEmailClient logs if emails are not sent.

### How to review

- look at file "dataset_ingress_v1.py" and check if anymore logging statements about sent emails should need removing.
- check that the logging statement under NopEmailClient is accurately describing what is happening and makes sense

### Who can review

Anyone


Moasib-Arif (approved):
LGTM

PR #147, opened 2024-06-07
830 implement str to bool

Moasib-Arif:
### What

Moved str_to_bool function and tests to dpytools.
Updated the version Release

### How to review

Sanity Check

### Who can review

Anyone


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "dpypelines"
-version = "0.1.0"
+version = "0.1.1"
```
SarahJohnsonONS: I don't think this needs to be updated, leave as 0.1.0

SarahJohnsonONS (commented):
Comment added

SarahJohnsonONS (approved):
LGTM

PR #150, opened 2024-06-24
889 Logging HTTP information.

nimshi89:
### What

Added an optional  `response` argument to the logging functions. 
Link to list of expected values: https://github.com/ONSdigital/dp-standards/blob/main/LOGGING_STANDARDS.md
Added functions to support the logging of responses to the `utility.py`
Made the 'testing.py' so the method can be tested.

### How to review

runt the testing.py and check the logging

### Who can review

anyone


jimwashbrook-ons:
Is this still needed?

PR #153, opened 2024-07-16
Acceptance test for generic ingress added

SarahJohnsonONS:
### What

Added acceptance test for `generic_file_ingress_v1()`. New steps for reading XML and asserting length/content match.

### How to review

Confirm that the test passes and that assertions are sufficient.

### Who can review

Anyone.


MuazzamChaud (approved):
Looks good to me

PR #155, opened 2024-08-02
Refactor matching repetitive functions

NickPapONS:
### What

Created a new function, "get_matching_pattern" in matching.py that combines the functionality of the two previously existing functions in this file, as they did essentially the exact same thing but just on different strings. Any usage of the previous functions has been replaced with the new "get_matching_pattern", and the unit tests have been adjusted to remove the tests for the old functions and test the same things on the new function (also avoids test repetition).

### How to review

Run the tests, look at the new function and ensure it all looks right.

### Who can review

Anyone


Moasib-Arif (commented):
Code Looks Good to me. I just have a question regarding the test coverage for supplementary_distribution

dpypelines/pipeline/dataset_ingress_v1.py
```diff
@@ -176,7 +173,9 @@ def dataset_ingress_v1(files_dir: str, pipeline_config: dict):
 
     # Extract the patterns for supplementary distributions from the pipeline configuration
     try:
-        supp_dist_patterns = get_supplementary_distribution_patterns(pipeline_config)
+        supp_dist_patterns = get_matching_pattern(
```
Moasib-Arif, line 176: Could this be tested if the supplementary_distributions is missing in the pipline config dictionary

Moasib-Arif (approved):
LGTM 👍

PR #160, opened 2024-08-12
Refactor transform functions

NickPapONS:
### What

Refactored the three functions in transform.py into one function. The three functions used to to the same thing but on different target fields of the input config dictionary. The new function takes in an argument for what target field of the dictionary to work on, cutting down on the repetitive code in transform.py and also allowing us to remove some repetitive unit tests that were associated with the three old functions.

### How to review

Look through the changed transform.py code and see if it makes sense/if you agree with the approach I took. Run the tests and see if it all works.

### Who can review

Anyone


tests/pipelines/pipeline/shared/pipelineconfig/test_transform.py
```diff
@@ -109,25 +59,33 @@ def test_get_transform_kwargs():
...
-        "transform_inputs": {"^data.xml$": sdmx_sanity_check_v1},
+        "transform_kwargs": {},
         "required_files": [{"matches": "^data.xml$", "count": "1"}],
         "supplementary_distributions": [{"matches": "^data.xml$", "count": "1"}],
         "secondary_function": dataset_ingress_v1,
     }
     with pytest.raises(AssertionError) as err:
-        get_transform_kwargs(config)
-    assert "'transform_kwargs' not found in config dictionary" in str(err.value)
+        get_transform_details(config, "transform_inputs")
```
Moasib-Arif, line 80: Currently, the tests focus only on the missing transform_inputs. Should we consider expanding the test coverage to include cases where transform or transform_kwarg might be missing from the config dictionary?

Moasib-Arif (commented):
Code looks good to me. Just one suggestion for the tests and then it is ready to be merged.

tests/pipelines/pipeline/shared/pipelineconfig/test_transform.py
NickPapONS: That can definitely be done if we want to be thorough with coverage. It would be doing the same thing as this test but checking a different string in the AssertionError.

Moasib-Arif (approved):
LGTM Approved

Moasib-Arif (approved):

PR #161, opened 2024-08-13
Bug fix `get_commit_id()`

SarahJohnsonONS:
### What

Glue job was failing due to the Git CLI not being installed inside the Glue job runner. Used `GitPython` library to get the commit hash of the current head. Test added.

### How to review

Run one of the ingress functions with `DISABLE_NOTIFICATIONS` set to false, and check that the commit hash is included in the Slack notification. 

### Who can review

Anyone.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,4 @@ __pycache__
 .coverage
 behave_debug.py
 features/fixtures/data/
-data/*
+data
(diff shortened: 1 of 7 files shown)


Moasib-Arif (approved):
Code looks good to me. Tested the Get_commit_id function it returns the id in Slack 👍

PR #164, opened 2024-08-19
refactor v1

nimshi89:
### What

created a new file `generate_version_metadata.py` and moved `generate_version_metadata` function to it
changed the import at `v1`, `v2` and `v21` to use the new file

### How to review

check if the import are correct

### Who can review
anyone


NickPapONS (changes requested):
Looks good, just one thing we can clean up as a result of your changes:

In dpypelines/pipeline/shared/transforms/sdmx/generic/v21/prototype/v1.py we can now get rid of the duplicate "generate_versions_metadata" function starting at line 121 of the file (like you already did in the v20 version of v1.py)

NickPapONS (approved):
Nice one, approved!

PR #165, opened 2024-08-19
Added additionitional info to notifier

MuazzamChaud:
### What

Update dp-data-pipelines to add more information to the `failure()` and `success()` messages of the `PipelineNotifier` so that it updates the automated Slack message to include the source ID and, ideally, a timestamp for both when the processing began and when it was completed, and which environment, i.e sandbox/staging/production.

### How to review

Update the success/failure messages as above, then check that the automated Slack message contains the source ID, timestamps and environment.

### Who can review

Anyone


dpypelines/pipeline/shared/notification.py
```diff
@@ -45,19 +45,25 @@ def success(self): ...
...
         ), "Unable to find required environment variable to populate webhook_url argument"
         self.client = SlackMessenger(webhook_url)
         self.notification_postfix = os.environ.get("NOTIFICATION_POSTFIX", "")
+        self.process_start_time = process_start_time
+        self.environment = get_environment()
 
-    def failure(self):
-        msg = f":x: {self.notification_postfix}, commit ID: {get_commit_id()}".strip()
+    def failure(self, source_id=None):
+        current_time = time.time()
```
Moasib-Arif: To remove code duplication "current_time..." AND "proccess_end_time" in  failure and sucess method. I would suggest create a new method called get_process_time which encapsulates the logic. You can call this method in failure and success.

dpypelines/s3_tar_received.py
```diff
@@ -22,17 +19,7 @@ def start(s3_object_name: str):
...
-        de_notifier: BasePipelineNotifier = notifier_from_env_var_webhook(
-            "DE_SLACK_WEBHOOK"
-        )
-        logger.info(
-            "de_notifier successfully instantiated", data={"type": type(de_notifier)}
-        )
-    except Exception as err:
-        logger.error("Failed to instanite de_notifier.", err)
-        raise err
+    notifier = get_notifier()
```
Moasib-Arif, line 22: When instantiating the notifier, I suggest reverting back to the previous code with try and except format. The original code should function correctly as is, and that format with try and except makes it easier to debug any issues when the script runs on AWS Glue.

dpypelines/s3_tar_received.py
```diff
@@ -47,7 +34,7 @@ def start(s3_object_name: str):
             err,
             data={"s3_object_name": s3_object_name},
         )
-        de_notifier.failure()
+        notifier.failure()
```
Moasib-Arif, line 37: Consider using de_notifier instead of notifier to maintain consistency with the rest of the code, where DE_notifier is used, and when setting environment variables.

Moasib-Arif (commented):
Code looks good to me. I have tested the changes by sending a notification message to Slack and it works as intended. Just couple of suggestions and then it should be ready to merge.

Moasib-Arif (approved):
LGTM 👍

PR #167, opened 2024-08-28
DIS-1065-changed-Notifications

nimshi89:
### What

Changed the `DISABLE_NOTIFICATIONS ` value from False to True 

### How to review

check if the notifications work by default, and they should not.

### Who can review

anyone


NickPapONS (approved):
Everything works and notifications don't get sent by default, approved.

PR #168, opened 2024-08-28
Get secondary function new dictionary access

NickPapONS:
### What
Amended the way we access the secondary_function field of the config dictionary. It used to be accessed by directly indexing the dictionary, which doesn't fit our usual method of using a function. The function that accesses the dictionary was placed in utils.py

### How to review

See if the placement/logic of the new function makes sense and run the tests to see if they pass.

### Who can review
Anyone


nimshi89 (approved):
LGTM

PR #169, opened 2024-08-30
Dis-1024 Changing to dataclass

nimshi89:
### What

removed the import for `namedtuples`
and added a decorator dataclass to the class instead
### How to review

all checks should pass

### Who can review
anyone


NickPapONS (approved):
All tests pass, change seems to have not broken/altered any functionality so all good!

PR #170, opened 2024-09-03
added fixes and lint

JimBryant7:
### What

I have added in fixes that should fix the issue with the glue job failing. The fix is in the file dpypelines/pipeline/shared/utils.py, this may need looking into more in the future as I think there must be a better way to return the branch name and commit id (I'm not certain that this won't always clone the main branch and could return the incorrect environment).

I've added a KeyError to dpypelines/pipeline/configuration.py as this will give better feedback in the logs if a source_id does not match within the configuration dict.

### How to review

Mainly check for anything obvious. I think the code should be looked at in more detail in the future but for now we want to get a fix in to unblock some other work.

### Who can review

Anyone

### DIFF
--- a/dpypelines/pipeline/configuration.py
+++ b/dpypelines/pipeline/configuration.py
@@ -45,8 +45,15 @@ def get_pipeline_config(source_id: str) -> dict:
     """
     Get pipeline config details for the given source_id
     """
+    pipeline_config = (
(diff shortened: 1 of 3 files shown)


SarahJohnsonONS (approved):
LGTM

PR #173, opened 2024-09-19
Removed unused file count from config

Moasib-Arif:
### What

Removed unused file count from configuration as it was not being used (Configuration.py / data_ingress.py).
Updated the tests where count was being used.

### How to review

Sanity check/ Run the test

### Who can review

Anyone


NickPapONS (approved):
Looks good to me, all tests pass 👍

PR #175, opened 2024-10-03
DIS-1009 reconfigure env variables

nimshi89:
### What

aded a addition condition to the assertion to check if the `SKIP_DATA_UPLOAD` is set to false and if it is false then check for the `UPLOAD_SERVICE_URL` to have a value
Removed all references to `UPLOAD_SERVICE_S3_BUCKET ` in `features/environment.py`

### How to review

Set the `SKIP_DATA_UPLOAD` to True and not provide a `UPLOAD_SERVICE_URL` and the assertion should not raise an error
### Who can review

anyone

### DIFF
--- a/dpypelines/pipeline/dataset_ingress_v1.py
+++ b/dpypelines/pipeline/dataset_ingress_v1.py
@@ -106,6 +106,8 @@ def dataset_ingress_v1(files_dir: str, pipeline_config: dict):
                 upload_url is not None
             ), "UPLOAD_SERVICE_URL environment variable not set."
             logger.info("Got Upload Service URL", data={"upload_url": upload_url})
+        else:
+            upload_url = os.environ.get("UPLOAD_SERVICE_URL", None)
     except Exception as err:
         logger.error("Error occurred when getting Upload Service URL", err)
         de_notifier.failure()
--- a/dpypelines/pipeline/dataset_ingress_v1.py
+++ b/dpypelines/pipeline/dataset_ingress_v1.py
@@ -102,7 +102,7 @@ def dataset_ingress_v1(files_dir: str, pipeline_config: dict):
     try:
         upload_url = os.environ.get("UPLOAD_SERVICE_URL", None)
         assert (
-            upload_url is not None
+            upload_url is not None and os.environ["SKIP_DATA_UPLOAD"] is not False
         ), "UPLOAD_SERVICE_URL environment variable not set."
         logger.info("Got Upload Service URL", data={"upload_url": upload_url})
     except Exception as err:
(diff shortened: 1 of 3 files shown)


dpypelines/pipeline/dataset_ingress_v1.py
```diff
@@ -100,11 +100,14 @@ def dataset_ingress_v1(files_dir: str, pipeline_config: dict):
 
     # Get Upload Service URL from environment variable
     try:
-        upload_url = os.environ.get("UPLOAD_SERVICE_URL", None)
-        assert (
-            upload_url is not None
-        ), "UPLOAD_SERVICE_URL environment variable not set."
-        logger.info("Got Upload Service URL", data={"upload_url": upload_url})
+        if os.environ["SKIP_DATA_UPLOAD"] is False:
```
SarahJohnsonONS: You need to handle the case where `SKIP_DATA_UPLOAD` is not set as an env var - there's some code on L360-L376 that you could move up (which also converts str to bool). Then do something like:

```
if skip_data_upload is False:
    try:
        upload_url = os.environ.get("UPLOAD_SERVICE_URL", None)
        assert (upload_url is not None), "UPLOAD_SERVICE_URL environment variable not set"
        logger.info("Got Upload Service URL", data={"upload_url": upload_url})
    except Exception as err:
        logger.error("Error occurred when getting Upload Service URL", err)
        de_notifier.failure()
        raise err
```

I think this should also work for `generic_ingress_v1`

SarahJohnsonONS (commented):
Comment added

dpypelines/pipeline/dataset_ingress_v1.py
nimshi89: okay made the changes but added and else statement cause the `upload_url` would not be assigned. This way the solution would require much change

nimshi89: made the changes, I did add an else part in cause the `UploadServiceClient` is using the `upload_url`.

SarahJohnsonONS: But if `SKIP_DATA_UPLOAD` isn't False, there won't be anything to upload, so you don't need the else statement because you don't need to set `upload_url`. Also, at the moment, because you're not calling `str_to_bool` on `SKIP_DATA_UPLOAD`, if it's set, then it will always be True (because it's a non-empty string) - you need to move the code from L361-376 to above this try-except block.

nimshi89: Sorted :)

SarahJohnsonONS (approved):
LGTM

PR #188, opened 2024-12-19
New upload methods implemented

SarahJohnsonONS:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

### What

Replaced individual `upload_new_<filetype>()` methods with `upload_new()` which checks the file extension of the file to be uploaded and sets the correct mimetype in the HTTP request. Added a `MIMETYPES` dictionary which includes the most common formats (CSV, JSON, XML, XLSX, CSDB). Added a utility function to get the correct mimetype, which returns a default of `None`. Tested locally with all specified filetypes, and the files were added to the output bucket - I also downloaded them from the bucket to check that I could open them.

### Testing

- [x] Yes
- [ ] No

### Documentation

- [ ] Yes
- [x] No
Small change, no documentation needed

### Related issues

Jira ticket: https://jira.ons.gov.uk/browse/DIS-2177

### How to review

Set env vars `UPLOAD_SERVICE_URL=anything` and `SERVICE_TOKEN_FOR_UPLOAD=anything` and run `generic_file_ingress_v1` with a valid pipeline config dict and various filetypes. Add a breakpoint on L279 to inspect the mimetype being pass to `upload_new()`.

If we want to make this even more flexible, we could change the default value in `get_mimetype()` to `application/octet-stream`, but not sure about security implications.


NickPapONS (commented):
Looks good and appears to work for me, just needs a formatting on the inputs.

NickPapONS (approved):
Nice one 👍

PR #192, opened 2025-01-20
2386 Unnecessary logging and error handling removed

NickPapONS:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

### What

Removed some instances of logging and error handling in data_ingress_v1.py that could be considered unnecessary.

For the logging, I removed instances that seemed to state the obvious (e.g. confirming something happened) when it isn't a significant process and also there is not much data being logged.

For the error handling, I removed instances of except blocks where we were catching Exceptions and raising an error, but the function being run already contains error handling in its definition (so if there was an error when running the function, it should be caught there). In the case of the function `get_transform_details`, this function is run several times and each instance had separate error handling. So I removed these and placed the try/except block in the function definition itself to avoid repetition (it should work the exact same way).

### Testing

- [ ] Yes
- [x] No
Doesn't add any new functionality.

### Documentation

- [ ] Yes
- [x] No
Doesn't add any new functionality.

### Related issues

https://jira.ons.gov.uk/browse/DIS-2386

### How to review

Look through the removed errors/logging in the changes, and determine whether you think it is right to remove each one or not. I approached this with the mindset of removing more rather than less, with the hope anyone reviewing this won't have to look through the entire file to search for instances I may have missed/forgotten to remove.

### DIFF
--- a/dpypelines/pipeline/dataset_ingress_v1.py
+++ b/dpypelines/pipeline/dataset_ingress_v1.py
@@ -80,20 +80,11 @@ def dataset_ingress_v1(files_dir: str, pipeline_config: dict):
         raise err
 
     # Retrieve submitter email from manifest_dict
-    try:
-        submitter_email = get_submitter_email(manifest_dict)
-        logger.info(
-            "Retrieved submitter email",
-            data={"submitter_email": submitter_email},
-        )
-    except Exception as err:
-        logger.error(
-            "Failed to retrieve submitter email",
-            err,
-            data={"manifest_dict": manifest_dict},
-        )
-        de_notifier.failure()
-        raise err
+    submitter_email = get_submitter_email(manifest_dict)
+    logger.info(
+        "Retrieved submitter email",
+        data={"submitter_email": submitter_email},
+    )
 
     # Create email client from env var
     try:
@@ -159,7 +150,6 @@ def dataset_ingress_v1(files_dir: str, pipeline_config: dict):
                         logger.error("metadata.json is not parseable as JSON", err)
                         de_notifier.failure()
                         raise err
-            logger.info("File exists and is not empty", data={"file": file})
             email_content = successful_validation_email(file)
             email_client.send(
                 submitter_email, email_content.subject, email_content.message
@@ -186,10 +176,6 @@ def dataset_ingress_v1(files_dir: str, pipeline_config: dict):
         )
         de_notifier.failure()
         raise err
-    logger.info(
-        "skip_data_upload set from SKIP_DATA_UPLOAD env var",
-        data={"value": skip_data_upload},
-    )
 
     # Retrieve Upload Service URL from environment variable
     if not skip_data_upload:
@@ -319,20 +305,7 @@ def dataset_ingress_v1(files_dir: str, pipeline_config: dict):
 
     # Get the transform inputs from the pipeline_config and run the specified sanity checker for it
     input_file_paths = []
-    try:
-        transform_inputs = get_transform_details(pipeline_config, "transform_inputs")
-        logger.info(
-            "Retrieved transform inputs",
-            data={"transform_inputs": transform_inputs},
-        )
-    except Exception as err:
-        logger.error(
-            "Failed to retrieve transform inputs from pipeline config",
-            err,
-            data={"pipeline_config": pipeline_config},
-        )
-        de_notifier.failure()
-        raise err
+    transform_inputs = get_transform_details(pipeline_config, "transform_inputs")
 
     for pattern, sanity_checker in transform_inputs.items():
         try:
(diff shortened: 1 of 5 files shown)


dpypelines/pipeline/dataset_ingress_v1.py
```diff
@@ -80,20 +80,11 @@ def dataset_ingress_v1(files_dir: str, pipeline_config: dict):
...
-    except Exception as err:
-        logger.error(
-            "Failed to retrieve submitter email",
-            err,
-            data={"manifest_dict": manifest_dict},
-        )
-        de_notifier.failure()
-        raise err
+    submitter_email = get_submitter_email(manifest_dict)
+    logger.info(
```
emmaons: Let's get rid of  logging that the submitter email is retrieved. 

Also you can incorporate all the email chunks into one bit, so let's do:

try:
     get_submitter_email.....
     get_email_client....
     log(email client crreated and submitter email retrieved)
except e:
     log(failed to create email client or retrieve submitter email , e).....

dpypelines/pipeline/validate_ingest_files.py
```diff
@@ -1,19 +1,39 @@
...
+def metadata_json_is_parseable(filepath, give_error: Optional[bool]) -> bool:
+    """
+    Check that the given file can be loaded as json. If not, return False.
+    Optional variable allows an error to be raised instead.
+    """
     try:
         with open(filepath, "r") as is_files_dir_parseable:
             json.load(is_files_dir_parseable)
+            logger.info(
+                "metadata.json is parseable as JSON", data={"file": filepath}
```
emmaons: This successful log can go :P

dpypelines/pipeline/shared/pipelineconfig/transform.py
```diff
@@ -13,8 +13,13 @@ def get_transform_details(config: Dict, transform_field: str) -> Any:
             transform_field in config.keys()
         ), f"""'{transform_field}' not found in config dictionary:
         {json.dumps(config, indent=2, default=lambda x: str(x))}"""
-        transform_field_to_get = config[transform_field]
-        return transform_field_to_get
+        try:
+            transform_field_to_get = config[transform_field]
+            return transform_field_to_get
```
emmaons, line 18: Great idea to move some of the error handling into the functions!

dpypelines/pipeline/generic_file_ingress_v1.py
```diff
@@ -134,7 +134,7 @@ def generic_file_ingress_v1(files_dir: str, pipeline_config: dict):
 
             # Validate that metadata.json is parseable as JSON
             if "metadata.json" in filepath:
-                if metadata_json_is_parseable(filepath):
+                if metadata_json_is_parseable(filepath, False):
                     logger.info(
                         "metadata.json is parseable as JSON", data={"file": file}
```
emmaons: can go

dpypelines/pipeline/dataset_ingress_v1.py
```diff
@@ -109,87 +89,30 @@ def dataset_ingress_v1(files_dir: str, pipeline_config: dict):
...
-        de_notifier.failure()
-        raise err
-    logger.info(
-        "skip_data_upload set from SKIP_DATA_UPLOAD env var",
-        data={"value": skip_data_upload},
-    )
+    skip_data_upload = str_to_bool(skip_data_upload)
 
     # Retrieve Upload Service URL from environment variable
     if not skip_data_upload:
```
emmaons, line 111: remove the successful log    logger.info(
                "Retrieved Upload Service URL",
                data={"upload_url": upload_url},
            ) on line 124

dpypelines/pipeline/dataset_ingress_v1.py
```diff
@@ -109,87 +89,30 @@ def dataset_ingress_v1(files_dir: str, pipeline_config: dict):
...
-                submitter_email, email_content.subject, email_content.message
-            )
-        except Exception as err:
-            logger.error(
-                "Failed to validate input file",
-                err,
-                data={"local_store_dir": files_dir, "file": file},
-            )
+        if not local_store.has_lone_file_matching(file):
+            err = FileNotFoundError(f"{file} does not exist")
```
emmaons: I think we can actually get rid of this if statement, we are reading in a list of files from a dir and then iterating through and checking they exist - they should exist by definition!

dpypelines/pipeline/validate_ingest_files.py
```diff
@@ -1,19 +1,36 @@
...
+from typing import Optional
 
 from dpytools.logging.logger import DpLogger
 
 logger = DpLogger("data-ingress-pipelines")
 
 
-def file_size_0(filepath) -> bool:
-    return os.stat(filepath).st_size == 0
+def file_size_0(filepath, give_error: Optional[bool]) -> bool:
```
emmaons: Let's make give_error False as a default value for both helper functions

tests/pipelines/pipeline/shared/validation/test_validate_generic_file_ingress.py
```diff
@@ -15,7 +15,7 @@ def test_import_file_size_not_0():
     """
     test_file = Path(fixtures_files_dir / "test_validate_csv_data.csv")
 
-    assert file_size_0(test_file) is False
+    assert file_size_0(test_file, False) is False
 
 
 def test_import_file_size_not_0_error():
```
emmaons: Add tests for when give_error is True and confirm they assert the error!

dpypelines/pipeline/validate_ingest_files.py
```diff
@@ -1,19 +1,36 @@
...
+    """
+    Checks the given file to ensure it is not empty. If it is, return True.
+    Optional variable allows an error to be raised instead.
+    """
+    if give_error and os.stat(filepath).st_size == 0:
+        err = ValueError(f"'{filepath}' is empty")
+        logger.error("Input file is empty", err, data={"file": filepath})
```
emmaons: remove logger here

dpypelines/pipeline/dataset_ingress_v1.py
```diff
@@ -109,87 +89,25 @@ def dataset_ingress_v1(files_dir: str, pipeline_config: dict):
...
-        except Exception as err:
-            logger.error(
-                "Failed to validate input file",
-                err,
-                data={"local_store_dir": files_dir, "file": file},
-            )
-            de_notifier.failure()
-            raise err
+        # Make sure file is not empty
+        file_size_0(filepath, give_error=True)
```
emmaons, line 96: Sorry this one will actually need the error handler!

emmaons (approved):

PR #193, opened 2025-01-20
DIS-2383-error-handler

nimshi89:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

### What

Created a new `error_handler_module.py` file. added a new method to handle errors
Created a new test file for the new functions `test_error_handler.py`

### Testing

- [x] Yes
- [ ] No

### Documentation

- [ ] Yes
- [x] No
- [x] Not as part of this ticket. (Could be done at a later point)

### Related issues

[Provide links to any related issues.](https://jira.ons.gov.uk/secure/RapidBoard.jspa?rapidView=3910&projectKey=DIS&view=detail&selectedIssue=DIS-2383)

### How to review

Check if the method makes sense and the tests are adequate.

### DIFF
--- a/dpypelines/pipeline/shared/error_handler_module.py
+++ b/dpypelines/pipeline/shared/error_handler_module.py
@@ -1,5 +1,7 @@
-from dpytools.logging.logger import DpLogger
 from typing import Optional
+
+from dpytools.logging.logger import DpLogger
+
 from dpypelines.pipeline.shared.utils import get_email_client
 from dpypelines.pipeline.utils import get_notifier
 
--- a/dpypelines/pipeline/shared/error_handler_module.py
+++ b/dpypelines/pipeline/shared/error_handler_module.py
@@ -7,35 +7,41 @@ from dpypelines.pipeline.utils import get_notifier
 
 logger = DpLogger("data-ingress-pipelines")
 
-def error_handler(section: str, error: str, data: Optional[dict], 
-                 submitter_email : str, surpress_logs: bool = True,
-                 surpress_email: bool= True, surpress_notification: bool= True 
+
+def error_handler(
+    section: str,
+    error: str,
+    data: Optional[dict],
+    submitter_email: str,
+    surpress_logs: bool = True,
+    surpress_email: bool = True,
+    surpress_notification: bool = True,
 ):
     """
-        This funciton handles the errors.
-        
-        Arguments:
-        section (str): The section of the ETL pipeline where the error occurred.
-        error (str): The error message to log and process.
-        data (dict, optional): Additional data to include in the logs.
-        surpress_logs (bool): If True, logging is surpressed.
-        surpress_email (bool): If True, email notifications are surpressed.
-        surpress_notification (bool): If True, system notifications are surpressed.
-    
+    This funciton handles the errors.
+
+    Arguments:
+    section (str): The section of the ETL pipeline where the error occurred.
+    error (str): The error message to log and process.
+    data (dict, optional): Additional data to include in the logs.
+    surpress_logs (bool): If True, logging is surpressed.
+    surpress_email (bool): If True, email notifications are surpressed.
+    surpress_notification (bool): If True, system notifications are surpressed.
+
     """
 
-    #Log errors if the `surpress_logs is set to false
+    # Log errors if the `surpress_logs is set to false
     if surpress_logs:
         if data:
             logger.error(f"Error in section: {section} {error}", data=data)
         else:
             logger.error(f"Error in section: {section} {error}")
 
-    #Send email notification if `surpress_email` is set to false
+    # Send email notification if `surpress_email` is set to false
     if surpress_email:
-        send_error_email(section,error,submitter_email,data)
+        send_error_email(section, error, submitter_email, data)
 
-    #Send system notifiations if `surpress_notification` is set to false
+    # Send system notifiations if `surpress_notification` is set to false
     if surpress_notification:
         try:
             notifier = get_notifier()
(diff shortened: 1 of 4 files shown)


dpypelines/pipeline/shared/error_handler_module.py
```diff
@@ -0,0 +1,55 @@
...
+from dpytools.logging.logger import DpLogger
+
+from dpypelines.pipeline.shared.utils import get_email_client
+from dpypelines.pipeline.utils import get_notifier
+
+logger = DpLogger("data-ingress-pipelines")
+
+def error_handler(section: str, error: str, data: Optional[dict], 
+                 submitter_email : str, surpress_logs: bool,
+                 surpress_email: bool, surpress_notification: bool
```
emmaons: Can we have default values of surpress_email, surpress_logs and surpress_notification as False

tests/test_error_handler.py
```diff
@@ -0,0 +1,128 @@
...
+    """Testing that `error handler` raises a `TypeError` when argument/arguments are missing"""
+
+    with pytest.raises(TypeError):
+        error_handler(section='1.1')
+    
+def test_error_handler_success_with_data():
+    """Testing if all arguments provided the then the function works as intended."""
+
+    with patch("dpypelines.pipeline.shared.error_handler_module.logger") as mock_logger, \
+         patch("dpypelines.pipeline.shared.error_handler_module.get_email_client", return_value=MagicMock()) as mock_email_client, \
```
emmaons: Try using the pytest patch decorator (@patch) outside the function, instead of with patch inside the test function, i.e 
@patch("dpypelines.pipeline.shared.error_handler_module.logger")
@patch("another_path")
def test_error_handler_sucess_with....(mock_another_path ,mock_logger)

^^note here the order of how things are passed in are inverted

dpypelines/pipeline/shared/error_handler_module.py
```diff
@@ -0,0 +1,55 @@
...
+        error (str): The error message to log and process.
+        data (dict, optional): Additional data to include in the logs.
+        surpress_logs (bool): If True, logging is surpressed.
+        surpress_email (bool): If True, email notifications are surpressed.
+        surpress_notification (bool): If True, system notifications are surpressed.
+    
+    """
+
+    #Log errors if the `surpress_logs is set to false
+    if not surpress_logs:
```
emmaons: Change the 'surpress_*' logs to positive booleans, output_logs, send_email, notify where the default values are true. Then in the if statements replace 'if not' with if. (I think this was my bad for how I defined the ticket :) )

tests/test_error_handler.py
```diff
@@ -0,0 +1,128 @@
...
+
+        #Test email client usage
+        mock_email_client.return_value.send.assert_called_once_with(
+            'test@gmail.com',
+            'ETL Pipeline error has occurred in Section: 1.2.1',
+            "An error has occurred in section: 1.2.1 \n\nThis is a Test Error\n\n Additional Data: {'TestKey': 'Test value'}"
+        )
+
+        #Test notifier usage
+        mock_notifier.return_value.failure.asser_called_once()
```
emmaons: Typo here - why is this not failing running the tests? When you run locally what happens exactly?

tests/test_error_handler.py
emmaons: Also - if you wrap all the tests in a parent class, then you will only need to define the mocked functions as patches ONCE and pass them in to all the tests :)

emmaons (changes requested):
1. Make fmt hasn't been run as there are linting diffs in the files , please run make fmt on the code.
2. Also, currently our PR pipeline github action isn't failing on linting checks. It is running 'poetry run black black dpyplines --diff'. Find the code quality file in the .github folder and update the --diff to --check

dpypelines/pipeline/shared/error_handler_module.py
nimshi89: Made the changes and I hope I didn't misunderstand :D

tests/test_error_handler.py
nimshi89: The notifiers failure method is called once and this checks that.  So that line just ensure the failure method has been called when it needed to. When I was making the changes and forgot to set the boolians in the test data, the test  did fail cause it wasn't triggered.

nimshi89: I am not sure about the local part I just usually run the test in vscode then on git and if it both passes then I take it it is working :)

tests/test_error_handler.py
```diff
@@ -0,0 +1,143 @@
...
+
+        # Test email client usage
+        self.mock_email_client.return_value.send.assert_called_once_with(
+            "test@gmail.com",
+            "ETL Pipeline error has occurred in Section: 1.2.1",
+            "An error has occurred in section: 1.2.1 \n\nThis is a Test Error\n\n Additional Data: {'TestKey': 'Test value'}",
+        )
+
+        # Test notifier usage
+        self.mock_notifier.return_value.failure.asser_called_once()
```
emmaons: still a typo here! This should error?

dpypelines/pipeline/shared/error_handler_module.py
```diff
@@ -0,0 +1,64 @@
...
+logger = DpLogger("data-ingress-pipelines")
+
+
+def error_handler(
+    section: str,
+    error: str,
+    data: Optional[dict],
+    submitter_email: str,
+    surpress_logs: bool = True,
+    surpress_email: bool = True,
```
emmaons: Yes as per your slack question - these need their names changed to output_logs, send_email, notify :)

tests/test_error_handler.py
```diff
@@ -0,0 +1,128 @@
+from unittest.mock import patch
+import pytest
+from dpypelines.pipeline.shared.error_handler_module import (
+    error_handler,
+    send_error_email,
+)
+
+
```
emmaons, line 10: I cleaned up the patch here so you can see the example! Note that they are patched in the order: log, email, notify but must be passed into the functions in the opposite order, i.e def test_error_handler_fail(self, mock_notifier, mock_email_client, mock_logger):

Also each function within the class needs to have them imported regardless of whether they are used

nimshi89: Thank you very much :) I did change the variables so the ticket is ready for re-review.

tests/test_error_handler.py
nimshi89: This is to call the failure method has been called once to check if the error notification was triggered once.

dpypelines/pipeline/shared/error_handler_module.py
```diff
@@ -0,0 +1,64 @@
...
+):
+    """
+    This funciton handles the errors.
+
+    Arguments:
+    section (str): The section of the ETL pipeline where the error occurred.
+    error (str): The error message to log and process.
+    data (dict, optional): Additional data to include in the logs.
+    surpress_logs (bool): If True, logging is surpressed.
+    surpress_email (bool): If True, email notifications are surpressed.
```
emmaons: Update surpress to enable!

emmaons (approved):

PR #195, opened 2025-01-24
2449 replace error handling

NickPapONS:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

### What

Changed instances of error handling in dataset_ingress_v1.py to use the new error_handler function.

### Testing

- [ ] Yes
- [x] No
Error handler is already tested and there is no new functionality.

### Documentation

- [ ] Yes
- [x] No
Functionality not changed, just using a different function to achieve the same thing.

### Related issues

### How to review

Look at the instances of the error_handler() function being called and determine if it is used correctly, logging a suitable error with correct data.

### DIFF
--- a/dpypelines/pipeline/dataset_ingress_v1.py
+++ b/dpypelines/pipeline/dataset_ingress_v1.py
@@ -8,12 +8,9 @@ from dpytools.stores.directory.local import LocalDirectoryStore
 from dpytools.utilities.utilities import str_to_bool
 
 from dpypelines.pipeline.shared.email_templates import (
-    failed_file_upload_email,
-    required_file_not_found_email,
     submission_processed_email,
     successful_file_upload_email,
     successful_validation_email,
-    supplementary_distribution_not_found_email,
 )
 from dpypelines.pipeline.shared.error_handler_module import error_handler
 from dpypelines.pipeline.shared.pipelineconfig.matching import get_matching_pattern
--- a/dpypelines/pipeline/dataset_ingress_v1.py
+++ b/dpypelines/pipeline/dataset_ingress_v1.py
@@ -43,7 +43,7 @@ def dataset_ingress_v1(files_dir: str, pipeline_config: dict):
     # Create notifier from webhook env var
     de_notifier = get_notifier()
 
-    #Create boolean variables for error handler, to allow easy setting
+    # Create boolean variables for error handler, to allow easy setting
     enable_notification = True
     enable_logs = True
     enable_email = True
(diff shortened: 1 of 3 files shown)


dpypelines/pipeline/dataset_ingress_v1.py
```diff
@@ -58,20 +56,30 @@ def dataset_ingress_v1(files_dir: str, pipeline_config: dict):
...
     except Exception as err:
-        logger.error(
-            "Failed to create local data store from files directory",
-            err,
+        error_handler(
+            section="1.1",
+            error="Failed to create local data store from files directory",
             data={"files_directory": files_dir},
+            submitter_email="No submitter email acquired yet",
+            enable_email=False,
```
emmaons: If enable_email, enable_logs and enable_notifications are the same throughout the script can we define the booleans once and then pass in these variables to the functions :)

dpypelines/pipeline/dataset_ingress_v1.py
```diff
@@ -58,20 +56,30 @@ def dataset_ingress_v1(files_dir: str, pipeline_config: dict):
...
+            enable_logs=True,
+            enable_notification=True,
         )
-        de_notifier.failure()
         raise err
```
emmaons: Error handler should raise error so this shouldn't be needed? Double check the function does and add this in if not and remove all the raise errs

dpypelines/pipeline/dataset_ingress_v1.py
```diff
@@ -58,20 +56,30 @@ def dataset_ingress_v1(files_dir: str, pipeline_config: dict):
...
     try:
         manifest_dict = local_store.get_lone_matching_json_as_dict("manifest.json")
     except Exception as err:
-        logger.error("Failed to retrieve manifest.json", err)
-        de_notifier.failure()
+        error_handler(
+            section="1.1",
+            error="Failed to retrieve manifest.json",
+            data=None,
+            submitter_email="No submitter email acquired yet",
```
emmaons: Can you edit error_handler so it checks if there is a submitter_email present as well as  if enable_email:
i.e if submitter_email and enable_email:
And if there is no submitter email available please leave this as an empty string.

dpypelines/pipeline/dataset_ingress_v1.py
NickPapONS: I made the changes, looks good except one part - all the variables that are assigned in the try/except blocks now get labelled as possibly unbound, I suppose because there is no exception being raised directly from the except block. It doesn't break anything but just wondering if that's alright.

emmaons (approved):

PR #197, opened 2025-01-30
Dis 2423 implement get path for dataset api

SarahJohnsonONS:

- [ ] Refactor
- [x] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

### What

Implemented the `DatasetAPIClient.get_path()` method to check for the existence of the API endpoint. This is based on the assumption that the `dataset_id` value can be extracted from `metadata.json` using the `dcterms:identifier` predicate. Currently, `dataset_ingress_v1()` prints out the necessary action to be taken to add a new dataset/update an existing dataset based on the API response, but the print statement for a 404 response doesn't currently print, due to how the rror is handled in `BaseHttpClient._handle_request()`. I'm also not sure how we are going to handle the difference between a nonsense `dataset_id` (one that shouldn't exist within the API), and a valid `dataset_id` that doesn't exist.

Note - when I was getting everything ready for review, I found a bug in the step uploading files to the Upload Service, which I have also fixed.

### Testing

- [x] Yes
- [ ] No

### Documentation

- [ ] Yes
- [ ] No
- [x] Not as part of this ticket. (Could be done at a later point)

### Related issues

Jira ticket: https://jira.ons.gov.uk/browse/DIS-2423

### How to review

Run `dataset_ingress_v1()` with data/metadata/manifest files, changing the `dcterms:identifier` value in the metadata, and check that the correct responses are generated. For the valid identifier, `cpih` should work. I have test files if you need them.


Makefile
```diff
@@ -18,4 +18,18 @@ test: install ## Run pytest and check test coverage
...
 
 feature: install
-	poetry run behave
\ No newline at end of file
+	poetry run behave
+
+
+symlink:
+	# Switch to production version
+	sed -i '' 's|{ git = "https://github.com/ONSdigital/dp-python-tools.git", tag = "v0.5.0" }|{path = "../dp-python-tools",  develop = true}|' pyproject.toml
```
emmaons: update the version of dpytools here and below- I know this isn't idea but will need to manually update for now!

dpypelines/pipeline/dataset_ingress_v1.py
```diff
@@ -141,6 +143,22 @@ def dataset_ingress_v1(files_dir: str, pipeline_config: dict):
                 enable_notification=enable_notification,
             )
 
+        try:
+            dataset_api_url = os.environ.get("DATASET_API_URL", None)
+            assert (
```
emmaons, line 133: merge these try except blocks (upload_url and dataset_api_url)

dpypelines/pipeline/utils.py
```diff
@@ -48,5 +51,34 @@ def get_upload_client(upload_url):
         raise err
 
 
+def get_dataset_api_client(dataset_api_url: str, dataset_id: str) -> DatasetAPIClient:
+    # Create DatasetAPIClient from dataset_api_url and dataset_id
```
emmaons: Let's get rid of this, it's specific to dataset id and looks like it will only be used once - I think it's extra abstraction without added functionality!

dpypelines/pipeline/dataset_ingress_v1.py
```diff
@@ -309,6 +330,52 @@ def dataset_ingress_v1(files_dir: str, pipeline_config: dict):
...
+            )
+        except Exception as err:
+            logger.error("Failed to retrieve metadata.json", err)
+            de_notifier.failure()
+            raise err
+
+        # Get dataset_id from metadata and create DatasetAPIClient
+        # 2423 TODO This is based on the understanding that the dataset_id will be the value associated with the dcterms:identifier predicate in metadata.json
+        dataset_id = get_value_from_metadata(metadata, "dcterms:identifier")
+        dataset_api_client = get_dataset_api_client(dataset_api_url, dataset_id)
```
emmaons: Just call dataset_api_client = DatasetAPIClient(dataset_api_url, dataset_id) and put the whole thing in the try: except block below (see comment on get_dataset_api_client)

dpypelines/pipeline/dataset_ingress_v1.py
```diff
@@ -309,6 +330,52 @@ def dataset_ingress_v1(files_dir: str, pipeline_config: dict):
...
+
+        # Check that the Dataset API endpoint exists
+        try:
+            dataset_api_response = dataset_api_client.get_path()
+            # 2423 TODO do we need to handle the difference between a nonsense dataset_id (i.e. one that shouldn't exist) and a valid dataset_id that doesn't yet exist in the Dataset API? Or will this be done within the API?
+            if dataset_api_response.status_code == 404:
+                # 2423 TODO This doesn't actually print, because of how the error is handled in BaseHTTPClient._handle_request()
```
emmaons: Explain what you mean here? Might need to edit this as we will need the code to execute on a 404

dpypelines/pipeline/utils.py
```diff
@@ -48,5 +51,34 @@ def get_upload_client(upload_url):
...
+            data={"dataset_api_url": dataset_api_url, "dataset_id": dataset_id},
+        )
+        raise err
+
+
 def get_secondary_function(config_dict: dict):
     return config_dict["secondary_function"]
+
+
+def get_value_from_metadata(metadata: dict, key: str) -> Optional[str]:
```
emmaons, line 57: Can you check the codebase and see if there is already a function that gets a key from a dictionary with error handling? I have a feeling there is so might not need to make a new function (could be wrong though!)

emmaons (changes requested):

dpypelines/pipeline/utils.py
SarahJohnsonONS: We've got a few different utility functions for getting values from dictionaries, but not one that takes both the dict and the key to look for:

dpypelines.pipeline.utils.get_source_id()
dpypelines.pipeline.utils.get_secondary_function() (am assuming this will change once Nick has finished his ticket)
dpypelines.pipeline.shared.utils.get_submitter_email() (this also does email validation, but we could pull the dict stuff out of it)
dpypelines.pipeline.shared.utils.get_mimetype()

I can update the `get_value...` function to be `get_value_from_dict` and implement it. I was also wondering about pulling all of the utilities into one file - not entirely sure why we have 2 separate `utils` files.

dpypelines/pipeline/dataset_ingress_v1.py
SarahJohnsonONS: This is the section from `handle_request`:

```
try:
            response = requests.request(method, url, *args, **kwargs)
            response.raise_for_status()
            return response

        except HTTPError as http_err:
            logger.error(
                f"HTTP error occurred: {http_err} when sending a {method} request to {url} with headers {kwargs.get('headers')}",
                http_err,
                data={
                    "http_error": http_err,
                    "method": method,
                    "url": url,
                    "headers": kwargs.get("headers"),
                },
            )
            raise http_err
```
I think that because 404 is an HTTP error, it's raised here, and execution stops, so it never gets to the print statement in `dataset_ingress`. The `logger.error` message gets printed out - that includes the Florence token, because it's one of the headers - not sure if this is a good idea...

emmaons (approved):

PR #198, opened 2025-01-30
2415 Updated all the validation steps

Moasib-Arif:

- [x] Refactor
- [x] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

### What

For ticket 2415. I have removed all the validation from Data_Ingres.py and created individual functions which will handle validation in Validate_pipeline.py becuase before all the validation was happenning throughout the script, but now it happens in one clean validation step which checks we have everything we need and then goes on to perform the ETL logic

Created tests for all the validation functions

Updated the the acceptance tests.

### Testing

- [x] Yes
- [ ] No

### Documentation

- [ ] Yes
- [x] No

### Related issues
Check all the logic is correct.
Run all the unit tests.
Run all the acceptance tests.

### How to review

Anyone


dpypelines/pipeline/validate_pipeline.py
```diff
@@ -0,0 +1,95 @@
...
+    """Validate and parse JSON file."""
+    try:
+        with open(file_path) as f:
+            data = json.load(f)
+        return data
+    except json.JSONDecodeError as e:
+        raise ValueError(f"File is not valid JSON: {str(e)}")
+
+
+def validate_manifest_vars(manifest_dict: dict) -> None:
```
emmaons: Can we change required_keys = ["manifestVersion", "source_id", "fileAuthorEmail"] as being hardcoded to an argument of the function

tests/pipelines/pipeline/shared/validation/test_validate_pipeline.py
```diff
@@ -0,0 +1,226 @@
...
+    assert "supplementary_files" in result
+
+
+def test_validate_pipeline_files_data_missing():
+    """
+    Tests that `validate_pipeline_files()` raises FileNotFoundError if the data file is missing.
+    """
+    files = test_cases_base_dir / "invalid_no_data"
+    pipeline_config = {
+        "config_version": 1,
```
emmaons: pipeline_config is duplicated, you can define it once at the top of the script

tests/pipelines/pipeline/shared/validation/test_validate_pipeline.py
```diff
@@ -0,0 +1,226 @@
...
+            {"matches": "^data.csv$"},
+            {"matches": "^metadata.json$"},
+        ],
+        "supplementary_distributions": [{"matches": "^supplementary.txt$"}],
+    }
+    with pytest.raises(FileNotFoundError) as e:
+        validate_pipeline_files(files, pipeline_config)
+
+    assert (
+        "Required file not found: tests/fixtures/test-cases/dataset_ingress_v1/invalid_no_metadata/metadata.json"
```
emmaons, line 60: Questions about how this works- if we are pointing to an empty directory each time, how does it fail at a different point each time? 
Also if we aren't doing anything to the directory before we run the tests then surely we can just point to one directory for all of them?

emmaons (changes requested):
Amazing work! Broken down logically, the right amount of error handling and great test coverage. Few small comments to look at :)

tests/pipelines/pipeline/shared/validation/test_validate_pipeline.py
Moasib-Arif: The reason why it fails because each directory is missing a different required file. For example, the invalid_no_metadata directory is only missing the metadata file whereas other required files like data or manifest are still there becuase it is missing only the metadata file the test will fail on the missing meetadata test function. Similarly other scenarios will point to directories which are missing specific files, ensuring each test fails at a unique check.If we use a single directory missing multiple files it will not allow us to isolate different missing files, as the first missing resource might overshadow subsequent checks. Keeping separate folders ensures that each scenario is tested distinctly and fails at the correct point.

features/steps/dataset_ingress.py
```diff
@@ -17,7 +17,8 @@
         "transform_inputs": {"^data.xml$": sdmx_sanity_check_v1},
         "transform_kwargs": {},
         "required_files": [{"matches": "^data.xml$"}],
-        "supplementary_distributions": [{"matches": "^data.xml$"}],
+        "supplementary_distributions": [],
+        "secondary_function": dataset_ingress_v1,
```
emmaons, line 21: secondary_function no longer exists! Remove this key

emmaons (approved):

PR #199, opened 2025-01-30
Remove secondary function logic

NickPapONS:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [x] Documentation Update

### What

Consolidates the logic for the two transform functions (dataset_ingress_v1 and generic_file_ingress_v1) into one. This allows us to remove generic_file_ingress_v1 and avoid the duplication we previously had. Also updates the tests and the documentation to reflect this change.

### Testing

- [ ] Yes
- [x] No

### Documentation

- [x] Yes
- [ ] No

### Related issues

### How to review

Have a look at the dataset_ingress_v1.py file and confirm it makes sense to consolidate the two functions this way. Let me know if I missed anything.


dpypelines/pipeline/dataset_ingress_v1.py
```diff
@@ -218,6 +223,22 @@ def dataset_ingress_v1(files_dir: str, pipeline_config: dict):
...
+                        required_file
+                    )
+                    logger.info(
+                        "File to be uploaded retrieved",
+                        data={"file_path": required_file_path},
+                    )
+                except Exception as err:
+                    logger.error(
+                        "Failed to retrieve file to be uploaded",
+                        err,
```
emmaons: handle_error should be used here

dpypelines/pipeline/configuration.py
```diff
@@ -39,7 +38,7 @@
             {"matches": ".*metadata.json$"},
         ],
         "supplementary_distributions": {},
-        "secondary_function": generic_file_ingress_v1,
+        "secondary_function": dataset_ingress_v1,
```
emmaons: let's get rid of this key entirely! secondary_function as a concept now no longer needs to exist

emmaons (changes requested):
Couple extra bits to check / remove:

I believe (* but shout if I'm wrong) that there is a function called something like get_secondary_function, which is called in s3/ingress .py files. Remove this and call the data_ingress function instead. Also remove all instances of the key 'secondary_function' (check the lambda function in the infra package too) :)

dpypelines/pipeline/configuration.py
NickPapONS: I see now, makes sense 👍

emmaons (approved):

emmaons (dismissed):

emmaons (approved):

PR #209, opened 2025-03-05
2708 GitHub workflow for release

Moasib-Arif:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

### What

Add github workflow for release, Github workflow is triggered when branch is merging into sandbox. Github workflow will increment the version and tag that commit with new version and builds wheel file.

### Testing

- [ ] Yes
- [x] No

### Documentation

- [ ] Yes
- [x] No

### Related issues

### How to review
Sanity check if logic is correct in release.yml file.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,55 +1,57 @@
+# Description:
+#   This workflow uses python-semantic-release to fully automate semantic versioning.
+#   Based on commit messages it determines version bumps (patch/minor/major),
+#   updates the changelog and version number, tags the commit, builds the wheel and creates a GitHub release.
+
 name: Release on Merge to Sandbox
 
 on:
   push:
     branches:
       - sandbox
-
-permissions:
-  contents: write
-  packages: write
+  pull_request:
+    branches:
+      - sandbox
+  workflow_dispatch:
 
 jobs:
   release:
     runs-on: ubuntu-latest
+    if: github.ref == 'refs/heads/sandbox'
+    permissions:
+      contents: write
+      packages: write
 
     steps:
       - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Force Branch to Latest Commit
+        run: |
+          git reset --hard ${{ github.sha }}
 
       - name: Set up Python
         uses: actions/setup-python@v4
         with:
           python-version: '3.9'
 
-      - name: Install poetry
-        run: pip install poetry
-
-      - name: Install dependencies
-        run: poetry install
-
-      - name: Increment version
-        id: version
+      - name: Install Dependencies
         run: |
-          poetry version patch
-          echo "version=$(poetry version -s)" >> $GITHUB_ENV
+          pip install poetry python-semantic-release
+          poetry install
 
-      - name: Tag commit with new version
+      - name: Bump Version
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
-          git config --global user.name 'github-actions[bot]'
-          git config --global user.email 'github-actions[bot]@users.noreply.github.com'
-          git tag v${{ env.version }}
-          git push origin --tags
+          semantic-release version --no-ci --yes
+
+      - name: Publish Release
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          semantic-release publish --yes
 
       - name: Build wheel
         run: poetry build
-
-      - name: Create GitHub Release
-        uses: actions/create-release@v1.1.0
-        env:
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,55 @@
...
+      - name: Install poetry
+        run: pip install poetry
+
+      - name: Install dependencies
+        run: poetry install
+
+      - name: Increment version
+        id: version
+        run: |
+          poetry version patch
```
jimwashbrook-ons: Do we always want to do a patch version, or are there instances we might want to increment major/minor/etc.?

.github/workflows/release.yml
```diff
@@ -0,0 +1,55 @@
...
+          git config --global user.name 'github-actions[bot]'
+          git config --global user.email 'github-actions[bot]@users.noreply.github.com'
+          git tag v${{ env.version }}
+          git push origin --tags
+
+      - name: Build wheel
+        run: poetry build
+
+      - name: Create GitHub Release
+        uses: actions/create-release@v1.1.0
```
jimwashbrook-ons: It's generally preferred to pin to major versions of actions (i.e. `actions/create-release@v1.1.0`), but there have been cases where that's been an issue when GitHub (or whoever else has made the action) have done breaking changes without incrementing the major version.

Out of curiosity, is there a reason this is pinned to `1.1.0` instead of the latest (`1.1.4` I believe)?

.github/workflows/release.yml
```diff
@@ -0,0 +1,55 @@
+name: Release on Merge to Sandbox
+
+on:
+  push:
+    branches:
+      - sandbox
+
+permissions:
```
jimwashbrook-ons: There's only one job in the workflow so this is fine _currently_, but I'd generally recommend putting permissions like this within the particular job that actually needs them.

If another job were added to this workflow, then that job would _also_ have these permissions but would likely not need them.

jimwashbrook-ons (commented):
This is a good start mate, good job 👍 

I'd add a little bit of documentation somewhere that documents this process (nothing substantial - just "<Workflow> runs <Trigger> and does <Thing>").

Would be good as well to hear your thoughts on what the next steps would be, in particular how do you think the release into other environments would work?

jimwashbrook-ons (dismissed):

jimwashbrook-ons (commented):
Looking good buddy!

There's a couple of potential problems I can see though, but it's hard for me to say 100% without seeing this run.

- the [`version`](https://python-semantic-release.readthedocs.io/en/latest/commands.html#cmd-version) command not only finds out the new version, but it also creates a commit with the new version tag, and can update things like `pyproject.toml` with the new version etc.
- As a result, with how this action is currently implemented, I _think_ what's going to happen is:

- Step **Run Semantic Release**: runs [`semantic-release publish`](https://python-semantic-release.readthedocs.io/en/latest/commands.html#semantic-release-publish).
  - Since no tag is specified, it will find the latest GitHub tag/release and publish that. 
- Step **Crete GitHub Release**: runs `semantic-release version` twice.
  - The first time this runs it will assumedly determine the next version, then increment the relevant things as configured (e.g. `pyproject.toml`), then create a new Git commit + tag with the new version, then push the commit+tag to GitHub
  - _Then_ it runs again, this time it will not do anything because there's no changes

I think we'll need change this so:
- We run `semantic-release version` first, and output the new version as a GitHub action output var
- Then we do the `semantic-release publish` bit
- Also don't think we need the GitHub release step at the end, because it _should_ be covered by the publish step

I think we'll also need to configure some things in the `pyproject.toml`, e.g. what branches this should trigger on etc. Otherwise it _probably_ won't detect that the `sandbox` branch is a valid branch for a version release

 FWIW there's a couple of pre-made GitHub Actions they've made that encapsulate some of this logic to make it easier, there's more detail available [on this doc page of theirs](https://python-semantic-release.readthedocs.io/en/latest/automatic-releases/github-actions.html). There's also an example full workflow at the bottom of that page which should cover the entire process (we also can simplify it as I don't think a couple of the earlier steps are necessary for us). You don't have to use them though but it might make things easier. They also don't cover configuration fully. 

Would be good as well, for development purposes, if you:
1. Amend the workflow to trigger on PRs to sandbox, and add a job condition so that it only runs if the branch is _this branch_
2. Add configuration for the Python Semantic Release so that a it releases from _this branch_, but releases from here are pre-release appended with "-wip" (or whatever - it doesn't matter).

That way, until we're 100% sure on this, we can ensure that the actual process is working as expected, and we can remove those extra bits before merging into sandbox.

jimwashbrook-ons:
I've made a few changes to this after testing it a fair bit, I'll fill you in Tuesday. Good work though :+1:

jimwashbrook-ons (approved):
👍

PR #214, opened 2025-03-11
Ref(dis 2590 zip)

nimshi89:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

### What

Describe what you have changed and why.

### Testing

- [ ] Yes
- [ ] No

### Documentation

- [ ] Yes
- [ ] No

### Related issues

### How to review

Describe the steps required to test the changes.


wip-integration-test.py
jimwashbrook-ons, line 1: I think this would be better either:
1. Removed
or
2. Moved into the tests dir

Maybe even a sub-directory in the tests directory? We could have something like

`./tests/functional-tests/s3_folder_received_test.py`

I like having these little functional test scripts myself as well, and it's always a shame to have to remove/recreate/etc them. But I think there's probably a better place for them than the root of the repo

dpypelines/pipeline/utils.py
```diff
@@ -148,19 +152,140 @@ def setup_clients():
...
+                data={"folder": str(dest_folder)},
+            )
+        shutil.move(str(src_folder), str(dest_folder))
+        logger.info(
+            "Moved extracted folder",
+            data={"folder": folder_name, "dest_folder": str(dest_folder)},
+        )
+    else:
+        err_msg = f"Expected folder '{folder_name}' not found in {src_dir}."
+        logger.error(err_msg)
```
SarahJohnsonONS: Missing `error` argument in `logger.error()`

dpypelines/pipeline/utils.py
```diff
@@ -148,19 +152,140 @@ def setup_clients():
...
+    it will be removed first.
+    """
+    folder_name = Path(zip_filename).stem  # e.g., 'e2e' from 'e2e.zip'
+    src_dir = Path(src_dir)
+    dest_dir = Path(dest_dir)
+    dest_dir.mkdir(parents=True, exist_ok=True)
+    src_folder = src_dir / folder_name
+    dest_folder = dest_dir / folder_name
+
+    if src_folder.exists() and src_folder.is_dir():
```
SarahJohnsonONS, line 262: `src_folder` (in this case `processing/integration_test`) doesn't exist (see other comment), so extracted files can't be moved into `dest_folder`. I'm also not sure what the advantage is of doing it as expressed here - why can't we just extract the zip file into a single folder (e.g. `unzipped`) and then create the `LocalDirectoryStore` from that folder, instead of extracting to `processing`, moving to `processed` and then creating the `LocalDirectoryStore`?

dpypelines/pipeline/utils.py
```diff
@@ -148,19 +152,140 @@ def setup_clients():
...
+    return local_zip_path
+
+
+def decompress_zip_file(zip_path: Path, dest_folder: Union[str, Path] = "processing"):
+    """
+    Decompress the given zip file into the specified destination folder.
+    """
+    dest_folder = Path(dest_folder)
+    dest_folder.mkdir(parents=True, exist_ok=True)
+    with zipfile.ZipFile(zip_path, "r") as zip_ref:
```
SarahJohnsonONS, line 227: This decompresses the zip file directly into `dest_folder`, not `dest_folder/<zip_file_stem>` so `move_extracted_folder()` fails because it can't find the files at the specified location.

SarahJohnsonONS (commented):
Comments added

dpypelines/pipeline/utils.py
```diff
@@ -148,19 +152,153 @@ def setup_clients():
...
+    Downloads a zip file from S3 into the 'input' folder and returns the local file path.
+    """
+    input_dir = Path("input")
+    input_dir.mkdir(parents=True, exist_ok=True)
+    zip_filename = os.path.basename(s3_object_name)  # e.g., e2e.zip
+    local_zip_path = input_dir / zip_filename
+
+    bucket_name = s3_object_name.split("/")[0]
+    object_key = "/".join(s3_object_name.split("/")[1:])
+    profile_name = "dp-sandbox"
```
jimwashbrook-ons: `os.environ.get("AWS_DEFAULT_PROFILE")`

jimwashbrook-ons (commented):
Couple of changes please mate

dpypelines/pipeline/utils.py
```diff
@@ -148,19 +152,181 @@ def setup_clients():
...
+
+    for file_path in extracted_folder.rglob("*"):
+        if (
+            file_path.is_file()
+            and str(file_path).endswith(".json")
+            or str(file_path).endswith(".csv")
+        ):
+            relative_path = file_path.relative_to("processing")
+            object_name = f"{bucket_name[0]}/processing/{relative_path.as_posix()}"
+            upload_local_file_to_s3(file_path, object_name, profile_name="dp-sandbox")
```
jimwashbrook-ons: Another hardcoded profile name

dpypelines/pipeline/utils.py
jimwashbrook-ons: Os is it AWS_PROFILE? 🤔

dpypelines/s3_folder_received.py
```diff
@@ -52,6 +56,29 @@ def start(s3_object_name: str, *args, **kwargs):
...
+        for file_path in extracted_folder.rglob("*"):
+            if (
+                file_path.is_file()
+                and str(file_path).endswith(".json")
+                or str(file_path).endswith(".csv")
+            ):
+                relative_path = file_path.relative_to("processed")
+                object_name = f"{bucket_name[0]}/processed/{relative_path.as_posix()}"
+                upload_local_file_to_s3(
+                    file_path, object_name, profile_name="dp-sandbox"
```
jimwashbrook-ons: And here

jimwashbrook-ons: Given there's 3 instance so fit, is it worth moving the profile name fetching elsewhere?

jimwashbrook-ons (dismissed):
Seems to work as far as I can tell, so good job there!

It doesn't seem to delete the `processing/` files on S3 (nor `input/`). We could also probably tidy this up a bit:

- Do we need to delete/move local files?
  - We can just extract _locally_
  - On the S3 bucket we can _move_ the .zip file on S3 to the processing folder (and then the processed folder)
- Would be good to try and move things from `dpypelines/pipeline/utils.py` into more appropriate places (I think some of it overlaps, or should be moved to` the `dp-python-tools` repository)
  - Good example of this is `download_zip_file`. This already exists, or something extremely close to it if not, in `dp-python-tools`.


For now I'm happy with this (once the hardcoded `dp-sandbox` variables are removed), and we'll clean this up in another PR.

Good job :+1:

jimwashbrook-ons:
Something like this:
image

Although there is actual good arguments for the extracted files being uploaded to the S3 (namely, being able to read the data via the Glue job etc), but for a MVP of the process I think the above is sufficient.

That being said, from this PR perpsective, I think just deleting the files from `processing/` after is the best change needed here :+1:

dpypelines/pipeline/utils.py
```diff
@@ -148,19 +152,183 @@ def setup_clients():
...
+    decompress_zip_file(local_zip_path, dest_folder="processing")
+
+    # Delete zip file
+    clean_directory("input")
+
+    extracted_folder = Path("processing") / local_zip_path.stem
+    bucket_name = s3_object_name.split("/")
+
+    for file_path in extracted_folder.rglob("*"):
+        if (
```
jimwashbrook-ons: Remove this `if` statement; if it's in the zip file let's just process it

jimwashbrook-ons (approved):
:+1:

PR #218, opened 2025-03-18
[2007] Start/End Logging for Future Performance Metrics

RedWalters:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

### What

Just a simple Start/End time logging that can be used in future for performance monitoring. 
Putting it in s3_folder_received.py with the intention that it can work for any future implemented pipelines.


jimwashbrook-ons:
Looks good to me, but what happens if there's an exception?

PR #248, opened 2025-04-23
Feat/2885/update pipeline to match new manifest and metadata specs

SarahJohnsonONS:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

### What

`dpypelines/pipeline/models.py`
Pydantic models created for Metadata and Manifest. I've set fields as `Optional` if they are not designated as `required` in `docs/schemas/openapi-specification.yaml`. I haven't put any custom validation in, as I'm not sure what needs validating beyond the basic validation provided through using Pydantic.

`dpypelines/pipeline/dataset_api.py`
Updated `get_post_request_values_from_metadata()` function to populate POST request from Pydantic Metadata model.

`dpypelines/pipeline/process_zip_file.py`
Functions related to processing zip files moved here from `dpypelines/pipeline/utils.py`.

`dpypelines/pipeline/validate_pipeline.py`
Validation simplified and updated to return Pydantic models of Metadata and Manifest instead of dicts. Functions removed related to `source_id` and `pipeline_config` as these are no longer compatible with pipeline functionality. I haven't yet deleted the `CONFIGURATION` dictionary or `get_pipeline_config()` function in `dpypelines/pipeline/configuration.py`, but probably should remove these?

`dpypelines/s3_folder_received.py`
Updated `start()` function following changes in code described above.

`tests/fixtures/test-cases`
Updated manifest/metadata test cases to match new format. Moved some things around to better match source code structure.

`tests/pipelines/*`
Moved some things around to better match source code structure.

### Testing

- [ ] Yes
- [x] No

### Documentation

- [ ] Yes
- [x] No

### Related issues

Jira ticket: https://jira.ons.gov.uk/browse/DIS-2885

### How to review

Check that the Pydantic models meet the requirements of the OpenAPI specification (`docs/schemas/openapi-specification.yaml`)


dpypelines/pipeline/dataset_api.py
```diff
@@ -8,41 +8,33 @@
...
-        edition_path = edition.get("dcterms:identifier", None)
-        metadata_distributions = edition.get("dcat:distribution", None)
-        if metadata_distributions is not None:
-            distributions = get_distribution_details_for_request(metadata_distributions)
-        else:
-            distributions = None
+        dataset_path = metadata.dataset_id
+        edition_path = metadata.edition
 
         request_body = {
```
jimwashbrook-ons: IMO:
1. Create a new base class for `Metadata` that has all the attributes except the things we're posting here (`dataset_id` and `edition`?)

image

2. You can then replace this big manual mapping with something simple like:
image

jimwashbrook-ons: I also think we can just get rid of this method entirely. Now that we're using classes, it's extremely obvious/simple/easy to get the dataset_path, edition_path, and request body.

dpypelines/pipeline/validate_pipeline.py
```diff
@@ -1,103 +1,38 @@
...
-        raise ValueError(f"File is not valid JSON: {str(e)}")
+def retrieve_and_validate_manifest(
+    local_store: LocalDirectoryStore,
+) -> Manifest:
+    """Retrieve and validate manifest from the local directory store."""
+    manifest_dict = local_store.get_lone_matching_json_as_dict("manifest.json")
 
+    if not manifest_dict:
+        err_msg = "Failed to retrieve manifest from the local directory store."
+        raise FileNotFoundError(err_msg)
```
jimwashbrook-ons, line 21: Doesn't `LocalDirectoryStore` already throw this error if `get_lone_matching_json_as_dict` fails?

dpypelines/pipeline/dataset_api.py
jimwashbrook-ons: And then we can simplify the `upload_metadata` method a bit, and then refactor it slightly to get something like
image

jimwashbrook-ons (dismissed):

dpypelines/pipeline/validate_pipeline.py
SarahJohnsonONS: If the json file is just an empty object (i.e the content is `{}`), `get_lone_matching_json_as_dict` returns an empty dict, but deosn't raise an error.

jimwashbrook-ons (approved):
Good work :+1:

PR #249, opened 2025-04-23
tests: Add integration tests

jimwashbrook-ons:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

#### Integration tests

Integration tests done where external services are mocked using pytest, and moto (for AWS). I believe this covers almost all, if not all, if the possible paths; if I've missed any please shout but let's cover that on a separate PR.

#### JobConfiguration

1. Modified to raise an exception in cases where:
  - The secret was not found in AWS Secrets Manager
  - A mapped key in the secret was not found in the secret
2. Amend how the default configs work (i.e. are returned from methods, not just global variables) for easier mocking

#### CI/CD - testing workflow

1. Amend the code coverage check to ensure that the code coverages is greater than or equal to 80%, not just greater than.
2. Add a step that runs the integration tests
 
#### Other

1. Couple of minor tweaks in places where `JobConfiguration` is used by another method/class to retrieve a variable, to simplify the code.
2. Make the string `"static"` its own object that is then referenced to, in the `check_dataset_type_is_static` method for best practices
3. Modify the `error_handler` method to raise the original exception thrown, not a new one. 
4. Modify the `makefile` so that:
  - Unit tests are a separate command
  - Integration tests are a separate command
  - The existing test command runs one and then the other [^1]

### Testing

Yes.

### Documentation

A minor README. If there's anything else y'all think would be appropriate please shout.

### Related issues

2835

[^1]: Running them all in one go causes problems because of the terrible idea to make `JobConfiguration` a singleton. This needs reworking, but for now the process actually works so this can be addressed in the future.

### DIFF
--- a/.github/workflows/testing.yaml
+++ b/.github/workflows/testing.yaml
@@ -13,7 +13,7 @@ permissions:
 
 jobs:
   unit-tests:
-    name: Pytest with code coverage
+    name: Unit tests with code coverage
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
@@ -40,7 +40,7 @@ jobs:
         run: poetry install
 
       - name: Pytest (make test)
-        run: poetry run pytest --cov=dpypelines --cov-report=xml
+        run: poetry run pytest --cov=dpypelines --cov-report=xml ./tests/pipelines
 
       - name: Get base branch commit
         id: get-base-ref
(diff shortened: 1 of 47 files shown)


SarahJohnsonONS (commented):
I've cross referenced with the user flow document, and the following scenarios aren't being covered that I can see - some of these may no longer be necessary:
- Unsupported data file format (i.e. not csv/xml/xls/xlsx/csdb)
- `edition` field missing from `metadata.json` (may not be necessary, as there's a test for missing `dataset_id`, and these are interrelated)
- Required field missing from `manifest.json`
- `manifest.json`/`metadata.json` are invalid JSON
- Invalid `submission_contact.email` in `manifest.json` (Pydantic will catch an incorrectly formatted email address, but do we need to deal with bouncebacks from addresses that don't exist?)

There's one other potential scenario that occurred to me while reviewing:
- No `current` object in the response to a `GET` request to endpont `/datasets/{dataset-id}` (see `dpypelines.pipeline.dataset_api.get_current_dataset_type()`)

One other thing I spotted - copy pasted comment `Test dataset type that isn't static` in `test_dataset_api_errors.py`, `test_job_configuration_errors.py` and `test_upload_service_errors.py`

NickPapONS (approved):
Approved 👍

PR #259, opened 2025-05-19
feat: Reduce RAM usage

jimwashbrook-ons:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

### What

1. Fixed CSV validator; it was printing every line :) Now only reads at most 2 lines. Much better.
2. Stopped uploading files to the S3. We're changing how this works soon anyway, and in the mean time it's just massively increasing the amount of RAM we use for no reason. This is causing incorrect testing results in the stress testing, so this should give more accurate results for future code
3. Fixing error handler tests. They were failing due to the errors being passed through weren't exceptions. No idea how these were working before

### Testing

- [ ] Yes
- [x] No

Updated existing unit tests.

### DIFF
--- a/dpypelines/pipeline/process_zip_file.py
+++ b/dpypelines/pipeline/process_zip_file.py
@@ -5,7 +5,7 @@ from pathlib import Path
 from typing import Tuple
 
 from dpytools.logging.logger import DpLogger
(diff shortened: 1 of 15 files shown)


NickPapONS (approved):
Looks good to me 👍

PR #261, opened 2025-05-20
docs: Add field to manifest for specifying that metadata is unchanged

jimwashbrook-ons:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [x] Documentation Update

### What

1. Add a field to the Manifest specification to signify that the metadata file for this version needs to be populated using the metadata from the _previous_ version
2. Make 2 new Metadata types that specify different required fields for clarity

### DIFF
--- a/docs/schemas/openapi-specification.yaml
+++ b/docs/schemas/openapi-specification.yaml
@@ -26,7 +26,10 @@ paths:
         content:
           application/json:
             schema:
-              $ref: "#/components/schemas/Metadata"
+              oneOf:
+                -  $ref: "#/components/schemas/NewMetadata"
+                -  $ref: "#/components/schemas/MetadataUpdates"
+
 components:
   schemas:
     UsageNote:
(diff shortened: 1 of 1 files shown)


docs/schemas/openapi-specification.yaml
```diff
@@ -138,13 +141,30 @@ components:
...
+    MetadataUpdates:
+      description: |
+        Metadata specification where the metadata from the _previous version_ is to be used for the new version, but values from _this metadata_ will override existing values.
+
+        `dataset_id` and `edition` are still required fields, as they will be needed to retrieve the metadata information from the Dataset API.
+      allOf:
+        - $ref: "#/components/schemas/Metadata"
+        - required:
+          - dataset_id
+          - edition
```
janderson2, line 167: Am I correct in assuming a new set of data files will be associated with this update? If so, then we will need to include the `distributions` as a required field as well listing these new data files.

Additionally, any new version that is created and published is required to have a `release_date`. Copying the previous version's release date wouldn't make sense here. Unless we intend to add the `release_date` later in the process then it will also be required.

janderson2 (changes requested):

docs/schemas/openapi-specification.yaml
jimwashbrook-ons: Very good points. Bit of a misunderstanding on my part here about what Tracker can validate in this instance. I've added those fields as required, as Niall believes that they should always be able to supply them.

janderson2 (approved):
LGTM

PR #271, opened 2025-05-29
feat: DatasetsCollection class

SarahJohnsonONS:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

### What

`DatasetsCollection` class with methods:
- `get_dataset` (by `dataset_id`)
- `get_all_datasets`
- `create_new_dataset`
- `update_dataset`

I also tweaked the `DatasetStatusesCollection.update_status` method to take an `event` argument, rather than the arguments to create an event, as there was a lot of repetition - use the `models.DatasetEventFactory` methods to create the relevant `event` and pass this as the `event` argument to `update_status`.

### Testing

- [x] Yes
- [ ] No

### Documentation

- [ ] Yes
- [ ] No

### Related issues

https://jira.ons.gov.uk/browse/DIS-3024


dpypelines/pipeline/db/db_collections.py
```diff
@@ -50,26 +58,32 @@ def get_all_statuses_for_dataset_id(self) -> List[models.DatasetStatus]:
...
 
         :param s3_object_key: The object key of the file received in the S3 ingest bucket
         :param additional_data: Any additional data associated with the event
 
         :return: models.DatasetStatus
         """
-
         # Create a `DatasetStatus` model with `status_type` of "PENDING"
         timestamp = dt.now()
+        file_name = s3_object_key.split("/")[-1]
```
jimwashbrook-ons: Move this to the factory

dpypelines/pipeline/db/db_collections.py
```diff
@@ -7,6 +7,14 @@
 import dpypelines.pipeline.models.db_models as models
 
 
+def _convert_ids_to_object_ids(dataset_document: Dict) -> Dict:
```
jimwashbrook-ons: Move this to the factories

dpypelines/pipeline/db/db_collections.py
```diff
@@ -166,3 +158,141 @@ def update_status(
         # Get updated status document and convert to DatasetStatus model
         updated_status_model = self.get_status(status_oid)
         return updated_status_model
+
+
+class DatasetsCollection:
```
jimwashbrook-ons: Move to another file and rename this one to be specific to the `DatasetStatusesCollection` class

dpypelines/pipeline/db/db_collections.py
```diff
@@ -23,7 +31,7 @@ def get_status(self, status_oid: ObjectId) -> models.DatasetStatus:
         # Get status document for the given ObjectID
         status_document = self.__collection.read_one_document({"_id": status_oid})
 
-        # Convert event id values to ObjectIds
+        # Convert event ID values to ObjectIds
         for event in status_document["events"]:
```
jimwashbrook-ons: Not from this PR but this overlaps with `_convert_ids_to_object_ids` and could be refactored

dpypelines/pipeline/db/db_collections.py
```diff
@@ -50,26 +58,32 @@ def get_all_statuses_for_dataset_id(self) -> List[models.DatasetStatus]:
         return all_status_models
 
     def create_new_status(
-        self, s3_object_key: str, additional_data: Optional[Dict[str, Any]] = None
+        self,
+        s3_object_key: str,
+        edition_id: Optional[str] = None,
+        version_id: Optional[str] = None,
```
jimwashbrook-ons: `version_id` is a nullable int on `DatasetStatus`

dpypelines/pipeline/db/db_collections.py
```diff
@@ -166,3 +158,141 @@ def update_status(
...
+        dataset_dict = dataset_model.dict_for_mongodb()
+        dataset_dict["_id"] = ObjectId(dataset_dict["id"])
+        self.__datasets_collection.create_one_document(dataset_dict)
+
+        return dataset_model
+
+    def update_dataset(
+        self,
+        dataset_id: str,
+        status_oid: str,
```
jimwashbrook-ons: All the other method args have `status_oid: ObjectId`; why is this one different?

dpypelines/pipeline/db/db_collections.py
```diff
@@ -166,3 +158,141 @@ def update_status(
...
+    def get_dataset(self, dataset_id: str) -> models.Dataset:
+        """
+        Get dataset from `datasets` collection and convert to Dataset model.
+
+        :param dataset_id: The dataset_id of the dataset document to retrieve
+
+        :return: models.Dataset
+        """
+        # Get dataset document for the given dataset_id
+        dataset_document = self.__datasets_collection.read_one_document(
```
jimwashbrook-ons: This comment applies elsehwere too but:

We've typed `DBCollection.read_one_document` as returning Dict. Actually it can return None (if none found).

We need to update the `DBCollection` typings appropriately, and handle not-found documents appropriately.

dpypelines/pipeline/db/db_collections.py
```diff
@@ -166,3 +158,141 @@ def update_status(
...
+        )
+
+        # Convert Dataset model to dict and insert into datasets collection
+        dataset_dict = dataset_model.dict_for_mongodb()
+        dataset_dict["_id"] = ObjectId(dataset_dict["id"])
+        self.__datasets_collection.create_one_document(dataset_dict)
+
+        return dataset_model
+
+    def update_dataset(
```
jimwashbrook-ons: This might be a good place for a transaction. How do you think we might implement that?

jimwashbrook-ons (commented):
Good work here buddy, few additional comments that are mostly not from this PR:

- Remove `dataset_id: str` from the constructor of the `DatasetStatusesCollection`, and move it to the appropriate methods

- We should modify `DocumentDBClient`:
    - Add `database_name` as constructor argument
    - Retrieve and store the database as a private local field
    - Remove the `db: Database` argument from the `get_collection` method

Reasoning for this is mainly that whilst we're mostly abstracting the `DocumentDb` logic away from everything, we're exposing `Database` completely. This isn't an ideal fix but since we're only using one database I think it's fine. This means that then we can:

- Modify the `DatasetStatusesCollection` and `DatasetCollection` to not take `COLLECTION: DBCollection` in the constructor, but the `DocumentDBClient`
    - Then the `DatasetStatusesCollection` and `DatasetCollection` can be responsible for initialising the `DBCollection` object from the `DocumentDBClient` instead
      - Probably a good place for an abstract class? Just a simple field for `collection_name: str`  and something like `def initialise_collection(self, client: DocumentDbClient)` which retrieves the collection from the client
  
I think these changes will then keep it all a bit cleaner otherwise we'll have something like:

```python
mongo_client = DocumentDBClient("localhost", 27017)
database = mongo_client.get_database("pipelines")
datasets_collection = mongo_client.get_collection(database, "datasets")
dataset_statuses_collection = mongo_client.get_collection(database, "dataset_statuses")

dataset_statuses = DatasetStatusesCollection(dataset_statuses_collection, "WHATEVER_DATASET_ID")
datasets = DatasetsCollection(datasets_collection, dataset_statuses) 
```

Which would become:

```python
mongo_client = DocumentDBClient("localhost", 27017, database="pipelines")
dataset_statuses = DatasetStatusesCollection(mongo_client)
datasets = DatasetsCollection(mongo_client, dataset_statuses) 
```

- `DatasetStatusesCollection.get_all_statuses_for_dataset_id` does a multiple database queries. Could we do this in one?

- I think `DatasetsCollection` should _solely_ focus on CRUD uperations for `Datasets`. I.e. I think it would be better to:
1. Remove the `DatasetStatusesCollection` from it, and all `DatasetStatus` operations it performs
2. Create _another class_, let's say `DatasetsService` (typical), that would then do what `DatasetsCollection` is _currently_ doing. For example (please don't take this too literally):
    - `create_new_dataset`
      - Creates the `Dataset` and initial `DatasetStatus` documents, saves to DB, etc
    - `update_dataset_status`
      - Creates/updates the `DatasetStatus` and updates the `Dataset` document

Benefits:
- Classes are now solely tied to their single responsibility
- Data access layers separate (and basically a [repository pattern](https://medium.com/@pererikbergman/repository-design-pattern-e28c0f3e4a30))
- Business logic is separate (managed by `DatasetService`). Can maybe implement transactions here to get a proper [unit of work](https://en.wikipedia.org/wiki/Unit_of_work) pattern.
- Easier to test
- Etc


Last thing:
- The model typings, and method argument typings, should be changed where you have non-None default values, but have typed the field as Optional

E.g.

```
retry_count: Optional[int] = 0
```

These are not actually optional since we've defined a default value. So `retry_count` will never be `None`; it'll either have a value set or it'll default to 0.

Hope this doesn't come off as critical! You've done really good work here and you should be happy with it. Now let's just make it 💯 🔥

dpypelines/pipeline/db/base_collection.py
```diff
@@ -0,0 +1,12 @@
...
+from dpytools.db.db_collection import DBCollection
+from dpytools.db.documentdb_client import DocumentDBClient
+
+
+class BaseCollection(ABC):
+    def __init__(self, collection_name: str):
+        self.collection_name = collection_name
+
+    @abstractmethod
+    def initialise_collection(self, client: DocumentDBClient) -> DBCollection:
```
jimwashbrook-ons: Remove the client from the method signature, and move it to the constructor. Then make this method non-abstract

At the moment in the concrete classes you're not actually using the client in the method signature, but are using the one in the concrete class. E.g.

```python
    def initialise_collection(self, client: DocumentDBClient): #<---- this is not used
        self.__collection = super().initialise_collection(self.client)
        return self.__collection
```

So if we change BaseCollection to this

```python
class BaseCollection(ABC):
    def __init__(self, collection_name: str, client: DocumentDBClient):
        self.collection_name = collection_name
        self.client = client

    @abstractmethod
    def initialise_collection(self) -> DBCollection:
        return self.client.get_collection(self.collection_name)
```

Then there's no need for this method to be overridden in the concrete classes

dpypelines/pipeline/db/dataset_statuses_collection.py
```diff
@@ -0,0 +1,177 @@
...
+
+        :param s3_object_key: The object key of the file received in the S3 ingest bucket
+        :param additional_data: Any additional data associated with the event
+
+        :return: models.DatasetStatus
+        """
+        # Create a `DatasetStatus` model with `status_type` of "PENDING"
+        timestamp = dt.now()
+        status_model = DatasetStatusFactory.create_dataset_status(
+            _id=ObjectId(),
```
jimwashbrook-ons: IMO move all the `_id=ObjectID()` to the factory methods.

dpypelines/pipeline/db/dataset_statuses_collection.py
```diff
@@ -0,0 +1,177 @@
...
+            dataset_id=dataset_id,
+            created_at=timestamp,
```
jimwashbrook-ons: And the timestamps, and anything that is always going to be _the same_. So `status` as well

dpypelines/pipeline/db/dataset_statuses_collection.py
```diff
@@ -0,0 +1,177 @@
...
+                "events": [
+                    event.dict_for_mongodb() for event in current_status_model.events
+                ],
+                "error_message": event.error_message,
+                "retry_count": event.retry_count,
+                "last_retry_timestamp": event.last_retry_timestamp,
+            }
+
+            # Handle upload events
+            if (
```
jimwashbrook-ons: I'd recommend splitting up some of this logic into separate methods, maybe with some of it on the `DatasetEvent` class.

Something like

```python
class DatasetEvent:
    # existing code goes here
    
    def is_upload_event(self) -> bool:
        return self.event_data and self.event_data.upload_location
    
class DatasetStatusesCollection:
    
    def update_status(self):
        #etc
        
        if upload_event and event.is_upload_event():
            __handle_update_event(event)
            
    def __handle_update_event(self, event, update_values):
        match event.event_data.upload_location:
            case models.UploadLocation.DATASET_API:
                update_values["uploaded_to_dataset_api"] = True
            case models.UploadLocation.UPLOAD_SERVICE:
                update_values["uploaded_to_upload_service"] = True
```

But, yknow, whatever floats your boat.

Bonus points if you do put something on `DatasetEvent`, and it's a `get` only property.

dpypelines/pipeline/db/dataset_statuses_collection.py
```diff
@@ -0,0 +1,177 @@
...
+                == models.UploadLocation.UPLOAD_SERVICE
+            ):
+                update_values["uploaded_to_upload_service"] = True
+
+            # Update status document with new values
+            self.__collection.update_one_document({"_id": status_oid}, update_values)
+
+        # Get updated status document and convert to DatasetStatus model
+        updated_status_model = self.get_status(status_oid)
+        if updated_status_model:
```
jimwashbrook-ons: Maybe raise an error here if not?

dpypelines/pipeline/db/datasets_collection.py
```diff
@@ -0,0 +1,72 @@
...
+        :param dataset_id: The dataset_id of the dataset document to retrieve
+
+        :return: models.Dataset
+        """
+        # Get dataset document for the given dataset_id
+        dataset_document = self.__collection.read_one_document(
+            {"dataset_id": dataset_id}
+        )
+
+        if dataset_document is not None:
```
jimwashbrook-ons: Would recommend inverting the if statement here, and raising an Exception. So

```python

if dataset_document is None:
    # Create this, make it take a type (e.g. "dataset") and an ID or something. Or the filters used to find it. Use your best judgement.
    raise DocumentNotFoundException()

dataset_document = _#ETC. rest of code
```

Generally better to exit earlier in cases like this so you don't have unnecessary nested statements which can make it harder to read :)

dpypelines/pipeline/db/datasets_collection.py
```diff
@@ -0,0 +1,72 @@
...
+    def get_all_datasets(self) -> Optional[List[models.Dataset]]:
+        """
+        Get all datasets from the `datasets` collection and convert to Dataset models.
+
+        :return: List[models.Dataset]
+        """
+        # Get all documents in the datasets collection
+        dataset_documents = self.__collection.read_many_documents({})
+
+        if dataset_documents:
```
jimwashbrook-ons: Same here with the exiting early. And there's a few other places so might be worth having a look through the rest to see where else you think it can be flattened out a bit.

dpypelines/pipeline/db/datasets_collection.py
```diff
@@ -0,0 +1,72 @@
...
+        return
+
+    def get_all_datasets(self) -> Optional[List[models.Dataset]]:
+        """
+        Get all datasets from the `datasets` collection and convert to Dataset models.
+
+        :return: List[models.Dataset]
+        """
+        # Get all documents in the datasets collection
+        dataset_documents = self.__collection.read_many_documents({})
```
jimwashbrook-ons: Is it worth making the argument for `self.__collection.read_many_documents`, which i assume is a dictionary for the filter query, optional? And then set it to `{}` there if the argument was `None`?

dpypelines/pipeline/db/datasets_service.py
```diff
@@ -0,0 +1,138 @@
...
+                self.datasets_collection.update_dataset(dataset_model, update_values)
+
+        except Exception as err:
+            raise err
+
+        # Get updated dataset document and convert to Dataset model
+        updated_dataset_model = self.datasets_collection.get_dataset(
+            dataset_id=dataset_id
+        )
+        if updated_dataset_model:
```
jimwashbrook-ons: And if not?

dpypelines/pipeline/db/datasets_service.py
```diff
@@ -0,0 +1,138 @@
...
+            return updated_dataset_model
+
+    def update_dataset_new_status(
+        self, dataset_id: str, new_status: models.DatasetStatus
+    ) -> Optional[models.Dataset]:
+        dataset_model = self.datasets_collection.get_dataset(dataset_id=dataset_id)
+
+        update_values = {
```
jimwashbrook-ons, line 134: This logic (`"statuses." + str(ID)`) is done in a couple of places; would be good to extract it out into a reusable method.

jimwashbrook-ons (dismissed):
Looks good to me mate, good job!

I think we should handle (or rather, throw) some exceptions in places where we're not getting data back where we're expecting. E.g. after we've updated a dataset, and we do a new query for the dataset, if we get nothing back then assumedly that's an error right?

And a few places I think could be tidied up slightly, and they're only minor adjustments so would be good to get in. And they're not _functional_ they're just best for readability, coding practices etc.

Oh and something I didn't put on the PR comments (sorry :D), but I think it would be good to have a factory class for creating the `DatasetsService`. It should create all the necessary related classes, then the `DatasetsService`, then return the `DatasetsService`.

E.g.

```python
class DatasetsServiceFactory:

@staticmethod
def create_datasets_service():
    db_client = ...
    datasets_collection = ....
    datasets_statuses_collection = ....
    datasets_service = ....
    return datasets_service
```

dpypelines/pipeline/db/dataset_statuses_collection.py
```diff
@@ -0,0 +1,177 @@
...
+            elif (
+                upload_event
+                and event.event_data
+                and event.event_data.upload_location
+                == models.UploadLocation.UPLOAD_SERVICE
+            ):
+                update_values["uploaded_to_upload_service"] = True
+
+            # Update status document with new values
+            self.__collection.update_one_document({"_id": status_oid}, update_values)
```
jimwashbrook-ons: There's a few places where we filter by `_id`, might be good to extract this to a small simple method in `BaseCollection`?

dpypelines/pipeline/db/db_models.py
```diff
@@ -0,0 +1,127 @@
...
+    FAILED = "FAILED"
+    COMPLETED = "COMPLETED"
+
+
+class UploadLocation(Enum):
+    DATASET_API = "DATASET_API"
+    UPLOAD_SERVICE = "UPLOAD_SERVICE"
+
+
+def _convert_status_ids_to_object_ids(dataset_document: Dict) -> Dict:
```
jimwashbrook-ons: Lot of overlap in this method with `_convert_event_ids_to_object_ids`. Can you refactor it any further? No worries if you can't. Just wondering what you think

dpypelines/pipeline/db/dataset_statuses_collection.py
```diff
@@ -0,0 +1,177 @@
...
+        :param status_oid: The ObjectID of the status document to retrieve
+
+        :return: models.DatasetStatus
+        """
+        # Get status document for the given ObjectID
+        status_document = self.__collection.read_one_document({"_id": status_oid})
+
+        if status_document:
+            # Convert event ID values to ObjectIds
+            status_document = _convert_event_ids_to_object_ids(status_document)
```
jimwashbrook-ons: This process is repeated in `get_all_statuses_for_dataset_id`, but for a collection of `DatasetStatus` documents.  Can we improve this?

dpypelines/pipeline/db/datasets_collection.py
```diff
@@ -0,0 +1,128 @@
...
+
+    def __init__(self, client: DocumentDBClient):
+        super().__init__(client, "datasets")
+        self.client = client
+
+    def initialise_collection(self):
+        self.__collection = super().initialise_collection()
+        return self.__collection
+
+    def get_dataset(self, dataset_id: str) -> Optional[models.Dataset]:
```
jimwashbrook-ons: I don't think this is Optional; it'll either return _a thing_, or it'll throw an exception

dpypelines/pipeline/db/datasets_collection.py
```diff
@@ -0,0 +1,128 @@
...
+        if not result.acknowledged:
+            raise DocumentNotUpdatedException(
+                message="Dataset document not updated",
+                data={"dataset_id": dataset_model.dataset_id},
+            )
+
+        # Get updated dataset document and convert to Dataset model
+        updated_dataset = self.get_dataset(dataset_model.dataset_id)
+
+        if not updated_dataset:
```
jimwashbrook-ons: Not needed

jimwashbrook-ons (dismissed):
:+1:

dpypelines/pipeline/db/dataset_statuses_collection.py
```diff
@@ -0,0 +1,211 @@
...
+        if not result.acknowledged:
+            raise DocumentNotUpdatedException(
+                message="Status document not updated in collection",
+                data={"_id": status_oid},
+            )
+
+        # Get updated status document and convert to DatasetStatus model
+        updated_status_model = self.get_status(status_oid)
+
+        if not updated_status_model:
```
jimwashbrook-ons: Not needed since `self.get_status` will error if None

dpypelines/pipeline/db/dataset_statuses_collection.py
```diff
@@ -0,0 +1,211 @@
...
+
+        return updated_status_model
+
+    def _handle_upload_event(
+        self, event: models.DatasetEvent, update_values: Dict[str, Any]
+    ) -> Dict[str, Any]:
+        match event.event_data.upload_location:  # type:ignore
+            case models.UploadLocation.DATASET_API:
+                update_values["uploaded_to_dataset_api"] = True
+            case models.UploadLocation.UPLOAD_SERVICE:
```
jimwashbrook-ons, line 198: No tests for this, can we add one please

dpypelines/pipeline/db/dataset_statuses_collection.py
```diff
@@ -0,0 +1,211 @@
...
+        :param event: The event associated with the status update
+        :param new_status: The status type associated with the update event
+        :param upload_event: Whether the update event is related to uploading to an external service (DatasetAPI or Upload Service)
+
+        :return: models.DatasetStatus
+        """
+        # Get current status
+        current_status_model = self.get_status(status_oid)
+
+        if not current_status_model:
```
jimwashbrook-ons: Not needed since `get_status` will return _something_ or error

dpypelines/pipeline/db/db_collection_factories.py
```diff
@@ -0,0 +1,21 @@
+from dpytools.db.documentdb_client import DocumentDBClient
+from dpypelines.pipeline.db.dataset_statuses_collection import DatasetStatusesCollection
+from dpypelines.pipeline.db.datasets_collection import DatasetsCollection
+from dpypelines.pipeline.db.datasets_service import DatasetsService
+
+
+class DatasetsServiceFactory:
+    @staticmethod
+    def create_datasets_service(client: DocumentDBClient) -> DatasetsService:
+        """
```
jimwashbrook-ons, line 10: Test please :)

jimwashbrook-ons (commented):
:100: 

LGTM good job

Few minor comments, only ones I care about are the unit tests though please :) 

Otherwise, there's a few things that are missing, or could be improved, but we'll cover that with the next step.

The main 2 things are:
1. Add logging to the various collection + service methods

2. Using pydantic to do things better, namely:
- The id <-> ObjectId conversion
- Serialising to MongoDB compatible dictionary better
- Use a proper class with pydantic validation for the s3_object_key parsing (and generaly S3 object event parsing/handling/etc)

jimwashbrook-ons (dismissed):

jimwashbrook-ons (approved):
🥇

PR #277, opened 2025-06-11
refactor: switch to pydantic_settings for job config

jimwashbrook-ons:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

#### Configuration changes

- Replaced custom `JobConfiguration` with `JobConfig`, which uses `pydantic-settings` to load secrets from AWS Secrets Manager + environment variables + dotnev files

1. Simpler
2. Easier
3. Allows overriding much easier

- Removed the singelton instance of `JobConfiguration`. Now we load the configuration at the start of the Lambda, and just pass it along as we go
    1. Better practice
    2. Much easier for testing

### Other

- Tried to shift more places to using dependency injection, instead of instantiating services in various places
    - Other than the `JobConfig` changes, I think the main place I've done this is with the `UploadServiceClient`. This is also now instantiated at the start of the Lambda, and passed to the method(s) that need it

- Reorganised various things, mainly from the original `utils.py`
    1. `upload_files` method now moved to `api/upload.py`, along with a new method for constructing the `UploadServiceClient`
    2. `send_submission_confirmation` method -> `messages/notification.py`
    3. `create_notifier` + `get_notifier` + `setup_clients` -> `messages/setup.py`

- Updated `requests` module to `2.32.4` to fix potential security issue, same as dpytools.

### Testing

- [x] Yes
- [ ] No

### Documentation

- [x] Yes
- [ ] No

### Related issues

N/A

### How to review

1. Run the Lambda locally like normal, ensure everything still works as expected
2. Create a `.env` file at the root of the repository, and overwrite some of the `JobConfig` variables there. Run the Lambda again and ensure the variables from .env take precedence
3. Do the same as 2 but also add an OS env variable, with the same name, and ensure _that_ takes predecedence over both 

The names must match either the actual property name, or the `alias` that has been set for them (which should just be the same but capitalised). E.g.:

```python
    dataset_api_url: str = Field(alias="DATASET_API_URL")
```

Here the value can be either `dataset_api_url` _or_ `DATASET_API_URL`


SarahJohnsonONS (approved):
Approved

PR #281, opened 2025-07-01
feat(3025): Code Cleanup and Simplification (State Management)

SarahJohnsonONS:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

### What

`ETLProcessor` class added (`dpypelines/pipeline/etl_processor.py`) with methods for performing the necessary ETL steps to extract a zip file from S3, perform validation and upload results to the Dataset API and Upload Service. This also incorporates the necessary database operations to track pipeline failures so that the retry lambda can be implemented.

### Testing

- [x] Yes
- [ ] No

### Documentation

- [x] Yes
- [ ] No

### Related issues

https://jira.ons.gov.uk/browse/DIS-3025

### How to review

Run the pipeline locally with a valid S3 object, and check that the database is updated with the correct records.


dpypelines/pipeline/db/global-bundle.pem
franmoore05, line 1: Should this file be encrypted or is it dummy data?

SarahJohnsonONS: It's the public key for AWS DocumentDB data encryption (https://docs.aws.amazon.com/documentdb/latest/developerguide/connect_programmatically.html#connect_programmatically-tls_enabled) so I don't think it needs to be encrypted?

franmoore05 (approved):

